PR #1, opened 2022-08-30
Clarify licensing terms

rosslagerwall:


MarkSymsCtx (approved):

lindig:
In some cases the new machine-readable license identifier is added and the old license removed. But with that any mention of Citrix is also removed. Is that as intended and how would that association be maintained or is is deliberately removed?

lindig (approved):

lindig:
@rosslagerwall can we merge this? But see my question above.

rosslagerwall:
> In some cases the new machine-readable license identifier is added and the old license removed. But with that any mention of Citrix is also removed. Is that as intended and how would that association be maintained or is it deliberately removed?

It was intentional:
1. The copyright years are typically not updated properly.
2. Once external contributions come in, the copyright notices would need to be updated with external contributor details which PRs typically don't do.

The code's copyright owners can be reconstructed by following git blame/history instead.

The only place the copyright lines were kept was with cpiofile.py which was copied from somewhere else and already had the copyright lines.

PR #2, opened 2022-09-15
Allow to install an host with IPv6 management interface

benjamreis:
Adds a screen to choose between IPv4, IPv6
and dual stack (with IPv4 as primary address type for management).

Write network conf file for IPv6 so that netinstall work in IPv6's modes

Configure the host with an IPv6 management interface

Signed-off-by: BenjiReis <benjamin.reis@vates.fr>


benjamreis:
Hi, any comments on this? :)

upgrade.py
```diff
@@ -487,6 +487,12 @@ def completeUpgrade(self, mounts, prev_install, target_disk, backup_partnum, log
             nfd.write("NETWORKING_IPV6=no\n")
             nfd.close()
             netutil.disable_ipv6_module(mounts["root"])
+        else:
+            # Enable IPV6
+            ipv6_conf = open("%s/etc/sysctl.d/91-net-ipv6.conf" % mounts["root"], "w")
```
lindig, line 492: As before, would prefer `with open as`

lindig (commented):
I think this looks good in general but I'm not familiar with this area; since it involves a UI change, could you include screenshots?

netinterface.py
```diff
@@ -32,8 +32,8 @@ class NetInterface:
     Autoconf = 3
 
     def __init__(self, mode, hwaddr, ipaddr=None, netmask=None, gateway=None,
```
MarkSymsCtx, line 34: all of the conditionals in this make this scream out for some heavy refactoring, splitting into different classes with static/dhcp helpers

netutil.py
```diff
@@ -236,15 +240,9 @@ def valid_vlan(vlan):
...
-    if not re.match('^\d+\.\d+\.\d+\.\d+$', addr):
-        return False
-    els = addr.split('.')
-    if len(els) != 4:
-        return False
-    for el in els:
-        if int(el) > 255:
-            return False
-    return True
+    ipv4_re = '^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}'
```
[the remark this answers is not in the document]
MarkSymsCtx: Why are we not using the ipaddress library here, is it just because we don't include it in the install image?

tui/installer/screens.py
```diff
@@ -941,7 +942,7 @@ def nsvalue(answers, id):
                 answers['manual-nameservers'][1].append(ns2_entry.value())
                 if ns3_entry.value() != '':
                     answers['manual-nameservers'][1].append(ns3_entry.value())
-            if 'net-admin-configuration' in answers and answers['net-admin-configuration'].isStatic():
+            if 'net-admin-configuration' in answers and answers['net-admin-configuration'].valid() and not answers['net-admin-configuration'].isDHCP():
```
MarkSymsCtx, line 945: This check is now complex enough to warrant pulling it out as a well named method.

tui/network.py
```diff
@@ -295,9 +493,12 @@ def specify_configuration(answers, txt, defaults):
             )
         ifaceName = conf_dict['config'].getInterfaceName(conf_dict['interface'])
         netutil.ifdown(ifaceName)
+        ifup_error = netutil.ifup(ifaceName) != 0
+        if conf_dict['config'].modev6 == NetInterface.Autoconf:
+            time.sleep(2)
 
         # check that we have *some* network:
-        if netutil.ifup(ifaceName) != 0 or not netutil.interfaceUp(ifaceName):
+        if ifup_error or not netutil.interfaceUp(ifaceName):
```
MarkSymsCtx, line 501: this (and the extraction of a variable) seems unnecessary.

tui/network.py
```diff
@@ -19,128 +19,326 @@
...
+        vlan_value = int(vlan_field.value()) if vlan_cb.selected() else None
+        if bool(dhcp_rb.selected()):
+            answers = NetInterface(NetInterface.DHCP, nic.hwaddr, vlan=vlan_value)
+        else:
+            answers = NetInterface(NetInterface.Static, nic.hwaddr, ip_field.value(),
+                                subnet_field.value(), gateway_field.value(),
+                                dns_field.value(), vlan=vlan_value)
+        return RIGHT_FORWARDS, answers
+
+    def get_ipv6_configuration(nic, txt, defaults, include_dns):
```
MarkSymsCtx, line 188: far too long and far too complicated

tui/network.py
benjamreis: The sleep is necessary to let some times for the autoconf of the interface.
It can be put inside the `netutil.ifup` method if you'd prefer.

tui/network.py
benjamreis: Not sure how to do better than a subfunction per screen.
TUI in code is very verbose.

netinterface.py
benjamreis: I agree, but I'd wager that this PR adds a new feature, to add refactoring on top of that would make it even more complex.
IMHO, a new PR (or several) should be created to refactor the code and perhaps adds some GH actions to enforce coding style and such.

MarkSymsCtx: Typically the process is to refactor before adding the new feature and refactor after all changes to remove duplication and complexity.

netutil.py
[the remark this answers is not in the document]
rosslagerwall: The code doesn't use `ipaddress` because it long predates its introduction in Python 3.3.

We could use one of the unofficial Python 2 backports but I'd probably prefer that we move to Python 3 first and then do further improvements.

In the meantime, I think using a regex here would be OK if the repeated patterns were condensed.

rosslagerwall:
Aside from the small comments above, I wonder if it would be possible to have a config option (or just a variable in constants.py) that controls whether the new configuration screens are shown? At least at this point, I don't think we'd want to expose this option to our users so it would be nice to have an easy way of hiding it.

backend.py
```diff
@@ -1502,15 +1502,16 @@ def configureNetworking(mounts, admin_iface, admin_bridge, admin_config, hn_conf
...
-            if manual_nameservers:
-                print >>mc, "DNS='%s'" % (','.join(nameservers),)
-            if domain:
-                print >>mc, "DOMAIN='%s'" % domain
         print >>mc, "MODEV6='%s'" % netinterface.NetInterface.getModeStr(admin_config.modev6)
         if admin_config.modev6 == netinterface.NetInterface.Static:
             print >>mc, "IPv6='%s'" % admin_config.ipv6addr
             if admin_config.ipv6_gateway:
                 print >>mc, "IPv6_GATEWAY='%s'" % admin_config.ipv6_gateway
+        if admin_config.valid() and not admin_config.isDHCP():
```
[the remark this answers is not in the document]
benjamreis: You're correct, the DHCP screen still allows to enter maual servers

tui/installer/screens.py
```diff
@@ -811,7 +811,8 @@ def ns_callback((enabled, )):
         for entry in [ns1_entry, ns2_entry, ns3_entry]:
             entry.setFlags(FLAG_DISABLED, enabled)
 
-    hide_rb = answers['net-admin-configuration'].isStatic()
+    admin_config = answers['net-admin-configuration']
+    hide_rb = admin_config.valid() and not admin_config.isDHCP()
```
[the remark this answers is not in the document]
benjamreis: Here it's to hide the radio button to select the auto DHCP nameservers so I think this is correct.
The code before my PR was already making a condition on the config being static.

tui/network.py
```diff
@@ -19,128 +19,326 @@
...
+        loop = True
+        while loop:
+            result = gf.run()
+
+            if buttons.buttonPressed(result) in ['ok', None]:
+                # validate input
+                msg = ''
+                if static_rb.selected():
+                    subnet_value = int(subnet_field.value())
+                    if not netutil.valid_ip_addr(ip_field.value()):
```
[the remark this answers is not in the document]
benjamreis: Yes the checks should be specific to an IP family

upgrade.py
```diff
@@ -487,6 +487,11 @@ def completeUpgrade(self, mounts, prev_install, target_disk, backup_partnum, log
             nfd.write("NETWORKING_IPV6=no\n")
             nfd.close()
             netutil.disable_ipv6_module(mounts["root"])
+        else:
```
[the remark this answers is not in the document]
benjamreis: `NETWORKING_IPV6` is written [here](https://github.com/xenserver/host-installer/pull/2/files#diff-0922596903b115becd729d4bd53e9ad6e2d5a70ca42034806e650d49db0b5cccR1549-R1559).

upgrade.py
```diff
@@ -487,6 +487,11 @@ def completeUpgrade(self, mounts, prev_install, target_disk, backup_partnum, log
...
+            # Enable IPV6
+            with open("%s/etc/sysctl.d/91-net-ipv6.conf" % mounts["root"], "w") as ipv6_conf:
+                for i in ['all', 'default']:
+                    ipv6_conf.write('net.ipv6.conf.%s.disable_ipv6=0\n' % i)
```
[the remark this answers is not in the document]
benjamreis: I don't think so, the IPv6 is working without doing it.

netinterface.py
benjamreis: Just realising: a netintrface can be configured for both IPv4 and IPv6 so splitting in 2 classes is not possible.

netutil.py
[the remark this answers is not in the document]
benjamreis: I removed the use of regex to instead use `socket.inet_pton`. Is that okay with you?

netinterface.py
benjamreis: I've created a `NetinterfaceV6` class, inheriting from `Netinterface`, to split the `__init__` codes between the 2 classes.

upgrade.py
```diff
@@ -487,6 +487,11 @@ def completeUpgrade(self, mounts, prev_install, target_disk, backup_partnum, log
             nfd.write("NETWORKING_IPV6=no\n")
             nfd.close()
             netutil.disable_ipv6_module(mounts["root"])
+        else:
+            # Enable IPV6
+            with open("%s/etc/sysctl.d/91-net-ipv6.conf" % mounts["root"], "w") as ipv6_conf:
```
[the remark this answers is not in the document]
benjamreis: I'm not sure to understand this comment. Could you point me towards this restore list?
I assume it allows to keep a config file upon an host upgrade.
Since this would be a new file is there something else to do?

benjamreis:
@rosslagerwall a quick question about the config option for enabling/disabling IPv6 support in the installer.
Is it still something wanted since it's now in the pipes to support IPv6 for XenServer as well AFAIK?

tui/network.py
benjamreis: I'v mutualized the codes for both IPv4 and IPv6 interface configuration form

benjamreis:
I fixed a few things, refactored the code to mutualize as much as possible.
Remains the questions of the restore list for `91-net-ipv6.conf` for which I need directions and if the param to enable the IPv6 config is still needed.

For the rest this is good to review, it has been tested.

netutil.py
```diff
@@ -137,7 +137,11 @@ def interfaceUp(interface):
     if rc != 0:
         return False
     inets = filter(lambda x: x.startswith("    inet "), out.split("\n"))
-    return len(inets) == 1
+    if len(inets) == 1:
+        return True
+
+    inet6s = filter(lambda x: x.startswith("    inet6 "), out.split("\n"))
+    return len(inet6s) > 1  # Not just the fe80:: address
```
lindig, line 144: This condition could be moved into the filter; this code is duplicated - so would it be worth to put it into a function?

lindig (approved):

netutil.py
lindig: I was thinking to filter the output such that only `inet6` that are not localhost remain. That would make it more explicit which ones you are looking for rather than assuming `fe80::` is in the list or results.

netinterface.py
```diff
@@ -184,20 +191,44 @@ def writeRHStyleInterface(self, iface):
...
+        iface_name = self.getInterfaceName(iface)
+        if self.isStatic4() and self.gateway and util.runCmd2(
+            ['/usr/sbin/arping', '-f', '-w', '120', '-I', iface_name, self.gateway]
+        ):
+            return False
+
+        if self.isStatic6() and self.ipv6_gateway and util.runCmd2(
+            ['/usr/sbin/ndisc6', '-1', '-w', '120', self.ipv6_gateway, iface_name]
+        ):
+            return False
```
benjamreis, line 229: This code is equivalent to the `arping` code above but for IPv6 as `arping` only supports IPv4.
Please note this requires to have `ndisc6` installed in host-installer image.

lindig: Is this reflected in the corresponding RPM spec file?

benjamreis: As far as I know your installer spec file repository is not public so I can't provide a PR to add `ndisc6` which is why I commented my code for someone with access to reflect the change. :)

netutil.py
benjamreis: So something like:
```py
inet6s = filter(lambda x: x.startswith("    inet6 ") and not x.startswith("    inet6 fe80:"), out.split("\n"))
```
?

lindig: Yes. Is there a risk that the formatting of the input changes such that it would be better to use `x.strip().find("inet6")` or similar?

benjamreis: I don't think so. More recent versions of `ipaddr` still use this format.

benjamreis:
Screens of the modified TUI:
Capture d’écran du 2023-06-21 09-34-19
management-ip-family-selector

lindig:
I don't know if this is pre-existing: the blue background for the text inside the pop-up make it really busy; I don't think these input elements need to be highlighted at all or a more subtle method should be used: text colouring or bold text but not changing background colour.

lindig:
@MarkSymsCtx commented that he would have liked to see a refactoring to avoid the heavy branching introduced by IPv6. While I think this is valid, I do like the counted work on this feature.

benjamreis:
> @MarkSymsCtx commented that he would have liked to see a refactoring to avoid the heavy branching introduced by IPv6. While I think this is valid, I do like the counted work on this feature.

I have splitted the PR in several commits to discriminated the refactorings and the changes of behavior to support IPv6 so reviewing commit per commit should allow to see what's what. :+1:

benjamreis:
> I don't know if this is pre-existing: the blue background for the text inside the pop-up make it really busy; I don't think these input elements need to be highlighted at all or a more subtle method should be used: text colouring or bold text but not changing background colour.

AFAIK this was already like this

bernhardkaindl (approved):

bernhardkaindl (approved):
I reviewed the "backend" changes. I did not review the frontend changes, but I guess that [IPv6 for xsconsole](https://github.com/xapi-project/xsconsole/pull/4) should be merged/integrated first and then both can be tested together.

I'll be away until about 16 August.

Not reviewed: [tui/network.py](https://github.com/xenserver/host-installer/pull/2/files#diff-9cb486eb8c76aeb0ed842ba9932fe10a026c14b980243c13c603e5ff3300bf39)

PR #4, opened 2022-10-18
init: allow 'answerfile' to work without 'install' boot parameter

ydirson:
Booting the installer with `answerfile=URL` and no `install` to
complement it would just get no effect, with just a cryptic "No
operation specified for answerfile" hidden in the logfile.

The `install` parameter had only one use in the current code: allowing
`answerfile` to be honored, so just ignore it.

### DIFF
--- a/doc/parameters.txt
+++ b/doc/parameters.txt
@@ -117,11 +117,6 @@ Startup
 Installer
 ---------
 
-  --install
-
-    Proceed with installation/upgrade.
-
-
   --answerfile=ans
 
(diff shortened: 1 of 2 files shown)


ydirson:
> Are we sure the flag that was removed is indeed unused? Looks good otherwise.

That's my reading, as far as the host-installer code is concerned.  Strictly speaking, though, there is still a possibility that the boot parameters would also be interpreted by another userspace program during installation.

rosslagerwall (approved):

lindig (approved):

PR #5, opened 2022-10-18
answerfile.txt: show proper cardinality of deprecated scripts

ydirson:


doc/answerfile.txt
```diff
@@ -256,8 +256,8 @@ Common Elements
 (Re)Install Attributes
 ----------------------
 
-  <installation sr-type="lvm|ext"?>
-  <installation srtype="lvm|ext"?>[D]
+  <installation sr-type="lvm|ext">?
```
stormi, line 259: I think the interrogation mark here was really meant to be next to the `sr-type` parameter, rather than next to the whole XML tag.

ydirson: right, dropping that line

doc/answerfile.txt
```diff
@@ -253,11 +253,11 @@ Common Elements
...
 
 
-(Re)Install Attributes
-----------------------
+(Re)Install Optional Attributes
+-------------------------------
 
-  <installation sr-type="lvm|ext"?>
-  <installation srtype="lvm|ext"?>[D]
+  sr-type="lvm|ext"
```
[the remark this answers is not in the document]
ydirson: OK, dropping that patch for now, needs much more thought that it looked on the surface

lindig:
Could you explain either here or in the commit message what this does and why the previous state was not proper? It's a bit cryptic because I am not familiar with the file format.

ydirson:
Those tags are listed without any "optional" qualifier, despite being deprecated.  A strict interpretation would have them seen as mandatory (which does not make sense given the deprecation).  I checked the code to see that only the first occurrence of those tags is honored, hence the `?` (would have been `*` if an arbitrary number of them was honored).

rosslagerwall (approved):

psafont (approved):

PR #6, opened 2022-10-24
answerfile.txt: document support for multiple <source> elements

ydirson:


doc/answerfile.txt
```diff
@@ -36,7 +37,7 @@ Restore:
 Common Elements
 ---------------
 
-  <source type="type">repo</source>
+  <source type="type">repo</source>+
```
rosslagerwall, line 40: `<source>` is not required for the `<restore>` mode so I think this should be `*`.

lindig (dismissed):
Just to help with the merge, approving it.

doc/answerfile.txt
ydirson: Hm, that makes things harder to document in a readable way :)
A `*` and an additional note could possibly work.

I'm not sure how much it would help people on the documenting side, but I had the thought of adding a DTD (old-school that I am...), which could at least help to validate an answerfile.  But then, with `<installation mode="upgrade">` not accepting the same content as the other `<installation>` modes, we would not be able to write a DTD that would reject all invalid files - we would need to use the additional expressiveness of an XML Schema.  Does that sound like a useful addition ?

While we're at that, I was also wondering how much of a good idea it to be to consider using a separate `<upgrade>` tag.  It seems different enough to justify it, and would probably make the doc easier to write and read, but if we want backward compatibility the implementation could suffer a bit in readability.

Which also brings me to the question of deprecation of old syntax: there are already some features documented as deprecated, and I did not see any explicit timeframe for their removal.  All of this should probably be separate discussions, do you think it's worth opening tickets to followup on them ?

doc/answerfile.txt
```diff
@@ -46,7 +46,8 @@ Common Elements
     nfs
 
   'source' defines the location of the installation repository or a
-  Supplemental Pack. There may be multiple 'source' elements.
+  Supplemental Pack. There may be multiple 'source' elements. It is
```
lindig, line 49: Would it be better to say: "The may be zero or multiple 'source' elements."?

ydirson: The problem is, for `<installation>` elements, it is mandatory, I believe insisting on the "zero" case which is only for `<restore>` can only add confusion.  I also realized `<bootloader>` and `<network-backend>`, also handled in `parseCommon()` are in the same situation: in fact the "Common elements" section relates to "Sub-elements for 'installation', common to all three modes".  So likely this section should be split in two, with `driver-source`, `fcoe-interface' and `*script` split as being the real "common elements".

ydirson:
Pushed a new version starting with the split of the section, and adding a 3rd patch documentating for an option I discovered while reading the source :)

lindig (approved):

rosslagerwall (approved):

rosslagerwall:
@ydirson It looks like this conflicts with one of your other commits that got merged. Could you please rebase this?

rosslagerwall (approved):

lindig (approved):

PR #15, opened 2022-12-07
Fix comment and range for dynamically-assigned device majors

ydirson:


diskutil.py
```diff
@@ -139,8 +139,8 @@ def getDiskList():
                 if major == 202 and isRemovable("/dev/" + name): # Ignore PV CDROM devices
                     continue
                 disks.append(name.replace("!", "/"))
-            # Handle LOCAL/EXPERIMENTAL and Block Extended Major devices
-            if 240 <= major <= 254 or major == 259:
+            # Handle dynamic and Block Extended Major devices
+            if 234 <= major <= 254 or 384 <= major <= 511 or major == 259:
```
rosslagerwall, line 143: Based on my reading of the documentation here: https://www.kernel.org/doc/Documentation/admin-guide/devices.txt

234-254 and 384-511 are for dynamic _char_ devices, not block devices so I don't think they should be included.
The 240-254 range for block devices is still called LOCAL/EXPERIMENTAL, not dynamic.

ydirson: damned, looks like I read to fast

ydirson:
Retracting

PR #19, opened 2022-12-12
util.runCmd2: log commands before running them

ydirson:


psafont (dismissed):

util.py
```diff
@@ -55,6 +55,7 @@ def copyFilesFromDir(sourcedir, dest):
 
 def runCmd2(command, with_stdout=False, with_stderr=False, inputtext=None):
 
+    logger.log("running %s" % str(command))
```
rosslagerwall, line 58: I think this might result in unnecessarily verbose logs - logging the command twice (before and after).

How about using a `try...except` and in the exception handler logging something like `Running {command} failed with exception {exception}` and then reraising the exception.

util.py
```diff
@@ -55,12 +55,16 @@ def copyFilesFromDir(sourcedir, dest):
...
-                           shell=isinstance(command, str),
-                           close_fds=True)
+    try:
+        cmd = subprocess.Popen(command, bufsize=1,
+                               stdin=(inputtext and subprocess.PIPE or None),
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE,
+                               shell=isinstance(command, str),
+                               close_fds=True)
+    except ex:
```
rosslagerwall, line 65: I think this should include the call to communicate() since that involves reading, writing, and waiting and it could fail for various reasons.

ydirson: Makes sense. I'd also get rid of the commented-out block in the middle, since we definitely don't want to use that buggy code (a `poll()` could do if needed, anyway), but likely in a preliminary patch, then.

psafont (dismissed):

util.py
```diff
@@ -55,28 +55,21 @@ def copyFilesFromDir(sourcedir, dest):
...
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE,
+                               shell=isinstance(command, str),
+                               close_fds=True)
+
+        # We could poll stdout/stderr for commands outputing large amounts
+        # of data, but the following should suffice in all cases
+        (out, err) = cmd.communicate(inputtext)
+        rv = cmd.returncode
+    except ex:
```
rosslagerwall, line 70: This is not valid syntax. It needs to be `except Exception as ex:`

ydirson: Good catch, a bit annoyed that python2 allowed me to write that :(

rosslagerwall (approved):

Issue #16, opened 2022-12-09
Improving behaviour with pre-existing software-RAID devices


ydirson:
Currently, udev rules in `/lib/udev/rules.d/65-md-incremental.rules` cause any pre-existing software-RAID devices to be auto-assembled, and the user cannot opt-out.  As a consequence:
* disks that are part of a RAID array of disks cannot be selected for install
* a RAID array built from partitions can be selected for install (but installed system won't boot)
* disks with partitions that are part of a RAID array are available for selection, but install will fail

To solve this situations, our current plans for XCP-ng are:
- disable auto-assembly (eg. by causing the `ANACONDA` udev envvar to be set), solving all the above problems
- still allow user to enable RAID when needed, identified use cases including:
  * being able to upgrade or restore a system that was installed in RAID device (we already have [support for such installs](https://github.com/xcp-ng/host-installer/commit/3656067c1c803552e2ae478227c6035861011e2a), not contributed yet, but I understand NetScaler has a use-case)
  * being able to install on a pre-existing RAID (would be the NetScaler use-case)

The basic idea for now is to probe the available disks for RAID superblocks (`mdadm --examine`), and when one is found, add an option besides the existing install/upgrade/restore/reinstall ones, to assemble any RAID volumes (`mdadm --assemble --scan`).  An option for this would be provided to activate this behavior from answerfile.

This will notably require moving the "Checking for existing products" logic later, so it can be rerun.  Its current location, before ensuring there are any disks in the system, already feels awkward; is there anything to be aware of around that area ?

rosslagerwall:
In general, I think what you're suggesting is fine.

But wouldn't it make sense to put the RAID stuff in the `F10 - set up advanced storage classes` option near the beginning of installation? After this point it is expected that all drivers are loaded and additional disks (e.g. FCoE) are present.

rosslagerwall:
I also think along with the IPv6 PR that this would ideally have a way of enabling/disabling the UI bits at build time so that we don't necessarily expose it to customers if we don't generally support it.

ydirson:
> But wouldn't it make sense to put the RAID stuff in the `F10 - set up advanced storage classes` option near the beginning of installation? After this point it is expected that all drivers are loaded and additional disks (e.g. FCoE) are present.

[...]

PR #20, opened 2022-12-12
tui: show the disk on which an existing installation or backup is found

ydirson:


psafont (approved):

MarkSymsCtx (approved):

tui/installer/screens.py
```diff
@@ -218,9 +218,11 @@ def get_admin_interface_configuration(answers):
 def get_installation_type(answers):
     entries = []
     for x in answers['upgradeable-products']:
-        entries.append(("Upgrade %s" % str(x), (x, x.settingsAvailable())))
```
rosslagerwall, line 221: I think the design goal was so that you don't need to see internal device names when installing upgrading/restoring. That way it looks cleaner and doesn't show (unstable) Linux device names. You can press F5 to see the full details of the existing installation, including the disk it is installed on. Does that help?

ydirson: Right, it does give the information.  However, the goal of this patch is to help a user avoid making wrong decisions without realizing it, and I'm not sure that relying on an optional display would help much.

An example use-case where it would help, very much tied to patch series we did not submit yet, is when a system has been installed on a software RAID device (we have in XCPng a patch allowing the user to build a RAID1 for use as primary disk). I'm working on a patch-series that makes RAID assembly "opt-in, off by default" (see #16), and after this behavior change such a RAID install will not be listed at first; then if any install is found on another disk, when the disk is not mentioned in the entry itself, the user might think the install listed is the one it expects to find, and I think we don't want that to happen.

So maybe I should just move this commit back into my RAID-improvements PR-to-be (which incidentally already exists as a [PR against XCPng](https://github.com/xcp-ng/host-installer/pull/6), and on which you're welcome to comment already), with better context.

rosslagerwall: OK, I see you have a reason for doing this. In that case I think it is generally fine to show some extra information there.

I'd like to ensure that it shows some kind of stable identifier rather than something like /dev/sda and /dev/sdb which might switch around on successive boots. Does it do this?

It might also be useful to include the disk size since that is often an easy way of disambiguating the disks in a system.

ydirson: I'm not really happy with the results on a small 80x25 textmode:
installer-longlines

With a longer description (faked here for testing), the screen overflow looks really bad:
installer-verylonglines

Unfortunately, that old newt library does not seem to support multiline list entries. Looks like for we'll have to be satisfied with just adding the size :/.

psafont (approved):

GeraldEV (approved):

ydirson:
@GeraldEV any reason to prefer rebase over merge?  Merge strategy would link the merge commit back to the PR, and also keeps the code in the exact same state that was tested by submitter (eg. making `git bisect` more useful).

PR #24, opened 2022-12-20
Keep `sched-gran`grub entry upon upgrade

benjamreis:
If a user set the `sched-gran` [through XAPI](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_host.mli#L524), the settings should be preserved after a host upgrade.

Signed-off-by: BenjiReis <benjamin.reis@vates.fr>


andyhhp:
This is probably ok for now, but I think we need a better way of identifying settings we don't wish to alter on upgrade.

GeraldEV (approved):

PR #27, opened 2022-12-22
Remove inconsistent "company name" from repartitioning screen.

ydirson:


tui/installer/screens.py
```diff
@@ -385,9 +385,9 @@ def repartition_existing(answers):
         "Convert Existing Installation",
         """The installer needs to change the disk layout of your existing installation.
 
-The conversion will replace all previous system image partitions to create the %s %s disk partition layout.
+The conversion will replace all previous system image partitions to create the %s disk partition layout.
 
-Continue with installation?""" % (COMPANY_NAME_SHORT, MY_PRODUCT_BRAND),
+Continue with installation?""" % (MY_PRODUCT_BRAND,),
```
stormi, line 390: Do you still need a tuple here?

ydirson: safety reflex measure, in case a variable used for formatting would change to contain a sequence (`%` accepting non-tuples is bad for this, only f-strings will save us, hopefully we'll have python3 there soon)

andyhhp: if it makes a difference, I'll insist on using the form that @ydirson provided.  Life is too short to deal with the fallout of the tuple/non-tupleness issues with `%`.

rosslagerwall (approved):

andyhhp (approved):

PR #35, opened 2023-01-16
snackutils: fix missing import

ydirson:


rosslagerwall (approved):

rosslagerwall:
We should also avoid importing * from snack at some point.

andyhhp (approved):

PR #37, opened 2023-01-23
LACP support

ydirson:
This branch deals with the installer side of things, firstboot data needs to be handled in XAPI, for which we depend on https://github.com/xapi-project/xen-api/pull/4333:
- configuration from cmdline (covers fetching answerfile and the next items)
- configuration from answerfile
- configuration UI for netinstall
- configuration UI of mgmt interface, writing data for firstboot

Known limitations we may want to address:
- bad cmdline config could be reported cleanly (we currently get a python backtrace, which gets garbled by the TUI)
- ~~does not deal with DebStyleInterface, ie. does not support Debian-based installer; shouldn't that code be simply be nuked ?~~ [nuked since]
- limitations that could be addressed by moving bond creation/management to a preliminar "advanced networking" step similar to "advanced storage" (adding a new F key for this to initial screen may look bad, though):
  - bond interface is not created immediately, so appears in the interface list as "[no link]". Maybe this can be dealt with, by first creating `ifcfg-*` files without an IP config.
  - cannot change our mind once the choice to create a bond with given `eth*` has been made

This PR explicitly does not deal with:
- creating more than 1 bond
- system upgrade parsing during upgrade (get mgmt-iface config in `ExistingInstallation.readSettings()`), to which `network_db` seems to be unable to report about bonding (but the impact seems to be limited to default values).  The system's bonds are already preserved on upgrade, though.

What the screens currently look like:

lacp-0-ifaces
lacp-1-create
lacp-2-confirm
lacp-3-ifaces


stormi:
What's the context of the last force-push?

ydirson:
> What's the context of the last force-push?

A simple rebase on latest release

PR #43, opened 2023-04-24
CP-42642: Restore NRPE service config

DeliZhangX:


upgrade.py
```diff
@@ -444,6 +444,9 @@ def buildRestoreList(self):
         self.restore_list += ['etc/mdadm.conf']
 
         self.restore_list += ['var/lib/xcp/verify_certificates']
+        
+        # NRPE service config
+        self.restore_list += ['etc/nagios/nrpe.cfg', 'etc/nrpe.d/xs.cfg']
```
andyhhp, line 449: surely we want to preserve everything in nrpe.d ?

rosslagerwall: In that case why is nrpe.cfg included here then? The design says `/etc/nagios/nrpe.cfg is non-customizable, so can replace directly` and as I understand it only `xs.cfg` will be modified.

DeliZhangX: Restore `/etc/nagios/nrpe.cfg` is just to avoid problems for customers, in case they modify some configuration in it. Backing up both files ensures integrity, otherwise customers might report case to support team.

andyhhp: that's also your answer to why you should restore the whole of nrpe.d

DeliZhangX: Sorry, we ensure integrity As Far As Possible and also keep it simple, other file than `/etc/nrpe.d/xs.cfg` is not supported explicitly (this will be mentioned in user document), also customer may add other config file but not put under `/etc/nrpe.d/`. In my previous design, I want to parse all 'include's and restore them, while reviewers said it's complex and need not, restore just the 2 supported files is enough.

DeliZhangX: Thank all, I have fixed to backup whole folder.

rosslagerwall (approved):

alexbrett (approved):

PR #44, opened 2023-05-17
Hide "Software RAID" button when not useful

ydirson:


ydirson:
Wrong PR target

stormi:
The PR target can be changed after the PR creation (but not the originating branch)

PR #45, opened 2023-05-19
CA-369899: NTP Server options, DHCP|Default|Manual|None

GeraldEV:
Clearing up time settings during install
New settings appear like this when using DHCP IP:
newoptions

Which leads to a new manual time setting screen like this:
nontp

The manual NTP server entry has also be stripped down to just entry of the servers (automatically enabled) - I forgot to grab a picture, let me know if you want to see it
New settings when using static IP:
staticoptions

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -2,6 +2,7 @@
 
 import os
 import os.path
+import stat
 import subprocess
 import datetime
 import re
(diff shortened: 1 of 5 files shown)


backend.py
```diff
@@ -493,9 +495,14 @@ def setTimeManually(localtime, set_time_dialog_dismissed, timezone):
     util.setLocalTime(timestr, timezone=timezone)
     assert util.runCmd2(['hwclock', '--utc', '--systohc']) == 0
 
-def configureNTP(mounts, ntp_servers):
+def setDHCPNTP(mounts, ntp_config_method):
+    script = os.path.join(mounts['root'], "etc/dhcp/dhclient.d/chrony.sh")
+    permission = "a+x" if ntp_config_method == "dhcp" else "a-x"
+    assert util.runCmd2(["chmod", permission, script]) == 0
```
rosslagerwall, line 501: It would be better to use `os.chmod` instead of shelling out to do this.

Otherwise, this looks good to me.

stormi:
Is this PR somewhat related to the fact that now that repodata and RPM signatures are verified during installation, a badly set clock causes signature verification failures? (which causes a simple "error installating package" error message)

GeraldEV:
@stormi no, this PR was due to a bug where the previous host installer options for NTP servers would also use the default CentOS NTP servers even after selecting the DHCP option (instead of just using the DHCP server)
Hence these improvements for finer and more accurate control over the NTP servers used after installation

PR #47, opened 2023-06-01
repository: remove unused "mount" parameter from getBranding

ydirson:
Stumbled on this while reading the code


bernhardkaindl (approved):
Thanks, the parameter 'mounts' is indeed not used.

BTW: argument vs parameter:
> When a function is called, the values that are passed during the call are called as arguments. The values which are defined at the time of the function prototype or definition of the function are called as parameters.


> A parameter is a variable in a function definition. It is a placeholder and hence does not have a concrete value. An argument is a value passed during function invocation. In a way, arguments fill in the place the parameters have held for them.

GeraldEV (approved):

Issue #46, opened 2023-05-30
is `DebStyleInterface` code still relevant?


benjamreis:
IIUC `DebStyleInterface` related code is dead and should be removed since now its the RH style that is used?
Am I correct in my assumption? Would you be okay if I open a PR to delete this code?

PR #48, opened 2023-06-02
Maintenance: remove dead code of DebStyleInterface

benjamreis:
Now only RH Style are used so this code can be removed 
Solves: https://github.com/xenserver/host-installer/issues/46


bernhardkaindl (approved):
From all I know, we can remove generating debian-style `/etc/network/interfaces` code as dead code.

PR #49, opened 2023-06-08
Drop depmod config files

ydirson:
This old depmod.conf is the default in el7, the fix was made upstream in rhel years ago.

Also drop its companion systemd hook to run `depmod -a`.

### DIFF
--- a/startup/depmod.conf
+++ b/startup/depmod.conf
@@ -1,1 +1,0 @@
-search updates extra built-in weak-updates


bernhardkaindl (dismissed):

startup/systemd-udevd_depmod.conf
```diff
@@ -1,2 +0,0 @@
-[Service]
-ExecStartPre=/sbin/depmod -a
```
rosslagerwall, line 2: This file is not here by accident. It is part of customization feature where it is possible to concatenate a CPIO archive containing a new or updated driver onto the end of the install.img. Running depmod before udev allows the new driver module to be found and loaded.

While I don't particularly like the feature, it might be better to leave this file as is unless there is a good reason to remove it?

ydirson: My main reason was that I thought it was oly there for overriding the depmod config, but in my tests I recall it did not seem to work anyway: when changing the depmod config, this was not sufficient to get the config taken into account, whereas when I launched `depmod -a` by hand it did work.

When digging for any logs that would complain about some unrecognized systemd directive or similar I was also under the impression that any such logs would not reach `/var/log` - which annoyed me quite a bit but I did not take the time to dig this yet.

rosslagerwall: I think `startup/depmod.conf ` is unnecessary as you said but `startup/systemd-udevd_depmod.conf` is unrelated and has a slightly arcane but valid use case.

ydirson: OK let's separate those issues then

GeraldEV (approved):

alexbrett (approved):

ydirson:
Or merged, unless someone objects?

PR #50, opened 2023-07-11
Update timezones

ukrbugreporter:
Change locale name to a more relevant one.
Source: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

### DIFF
--- a/timezones
+++ b/timezones
@@ -364,7 +364,7 @@ Europe/Isle_of_Man
 Europe/Istanbul
 Europe/Jersey
 Europe/Kaliningrad
-Europe/Kiev
+Europe/Kyiv
 Europe/Lisbon
 Europe/Ljubljana
 Europe/London


MarkSymsCtx (approved):

bernhardkaindl (changes requested):
While it would be politically preferable to use the Ukrainian word for the city for the timezone, I think it has to use what the Dom0 system understands, and so far I can't see Europe/Kyiv in the tzdata file of our current Dom0 rootfs:
```sh
grep -r Europe/K /usr/share/
/usr/share/zoneinfo/tzdata.zi:Z Europe/Kiev
```
Therefore, AFAIK, until we have more info we've to update to tzdata which understands Kyiv before we can merge this, unfortunately.

As far as I understand, this PR has to be blocked until a `tzdata` with it is available for XenServer/XCP for/in installations.

GeraldEV:
We will shortly be updating the host-installer to Python 3, this PR looks like it will remain open for some time and will need updating to be Python 3 compatible (if not already) once the upgrade changes have been made

MarkSymsCtx:
> I think the issue here is system compatibility. We still use CentOS 7 which has no "Kyiv". Unless we make sure we have it (for instance adding a symbolic link) this change will break on CentOS 7.

We should probably update tzdata anyway for general system sanity. It typically gets updated fairly regularly on standard Linux distributions and whilst we were originally derived fro CentOS 7 we have long ago diverged into our own distribution.

PR #52, opened 2023-07-17
Keep user multipath configuration upon upgrade

benjamreis:
All files matchin `custom.*\.conf` will be kept upon upgrade.

See: https://github.com/xapi-project/sm/pull/600

### DIFF
--- a/upgrade.py
+++ b/upgrade.py
@@ -683,6 +683,9 @@ class ThirdGenUpgrader(Upgrader):
         restore_list += ['etc/stunnel/xapi-pool-ca-bundle.pem', {'dir': 'etc/stunnel/certs-pool'}]
         restore_list += ['etc/stunnel/xapi-stunnel-ca-bundle.pem', {'dir': 'etc/stunnel/certs'}]
 
+        # Keep user multipath configuration
+        restore_list += [{'dir': 'etc/multipath/conf.d', 're': re.compile(r'custom.*\.conf')}]
+
         return restore_list
 
     completeUpgradeArgs = ['mounts', 'installation-to-overwrite', 'primary-disk', 'backup-partnum', 'logs-partnum', 'net-admin-interface', 'net-admin-bridge', 'net-admin-configuration']


upgrade.py
```diff
@@ -448,6 +448,9 @@ def buildRestoreList(self):
         # NRPE service config
         self.restore_list += ['etc/nagios/nrpe.cfg', {'dir': 'etc/nrpe.d'}]
 
+        # Keep user multipath configuration
+        self.restore_list += [{'dir': 'etc/multipath/conf.d'}]
```
stormi, line 452: Here we need to choose between either saving the whole directory, or just `custom.conf`. Saving the whole directory may be a perk when/if vendors start taking the habit of dropping a file of theirs into this directory when needed.

On the other hand, if we want to keep the liberty of packaging configuration files in `/etc/multipath/conf.d`, then it may be safer to just add `etc/multipath/conf.d/custom.conf` to the list.

benjamreis:
Hi, we reduced to a specific regex to only save custom config files to avoid conflic with RPM installed files.

What do you guys think? :)

alexbrett:
From the sm PR, it seems we state in multipath.conf:
```
# For custom multipath configuration, create a separate .conf file in the
# /etc/multipath/conf.d/ directory.
```

However, this change only preserves files that match `custom*.conf`, which is not quite the same. If people have found the `custom.conf` file they may see the text in that which says it'll be preserved, but if they have only looked at multipath.conf, and then created e.g. /etc/multipath/conf.d/my_new_file.conf, they might still expect that to be preserved based on the text in multipath.conf.

As such it seems to me the text in multipath.conf should either say e.g. `create a separate .conf file in the /etc/multipath/conf.d/ directory. Note filenames beginning with the name custom, will be preserved during system updates/upgrades` (or something worded slightly better to that effect), or we should preserve everything, otherwise we risk confusing people?

ydirson:
@alexbrett the question would be, would it be valid for eg. an OEM to install files there though a RPM?  That's a typical use for the "conf.d" idiom, besides letting user modifications not interfere with system updates.
If we consider that for this particular case it makes no sense (I'm myself not familiar enough with the matter), then fine.  But we could also decide to restrict the liberty given in `multipath.conf` (which incidently people may already have used, making this path maybe harder to choose...)

ydirson:
What do you decide?  Would you rather simply keep `*.conf`?

PR #56, opened 2023-07-28
CP-43949: Raise exception when we can't find networkd.db

GeraldEV:
This covers the case where we have a management interface (mgmt_iface) from /etc/xensource-inventory but we can't find networkd.db in the installation (in the case where we've started an install/upgrade but it was interrupted)

net-admin-bridge or net-admin-interface is required by performInstallation, the failure if they are missing is abstract and unhelpful, this change will fail earlier and more clearly

### DIFF
--- a/product.py
+++ b/product.py
@@ -228,6 +228,7 @@ class ExistingInstallation:
 
             if not mgmt_iface:
                 logger.log('No existing management interface found.')
+                raise SettingsNotAvailable("Could not find network configuration")
             elif os.path.exists(self.join_state_path(constants.NETWORK_DB)):
                 logger.log('Checking %s for management interface configuration' % constants.NETWORKD_DB)
 
@@ -280,6 +281,9 @@ class ExistingInstallation:
                     results['net-admin-configuration'].addIPv6(NetInterface.DHCP)
                 elif protov6 == 'autoconf':
                     results['net-admin-configuration'].addIPv6(NetInterface.Autoconf)
+            else:
+                logger.log("Failed to find " + self.join_state_path(constants.NETWORK_DB))
+                raise SettingsNotAvailable("Could not find network configuration")
 
             repo_list = []
             if os.path.exists(self.join_state_path(constants.INSTALLED_REPOS_DIR)):
--- a/tui/installer/__init__.py
+++ b/tui/installer/__init__.py
@@ -24,8 +24,11 @@ def need_networking(answers):
            answers['source-media'] in ['url', 'nfs']:
         return True
     if 'installation-to-overwrite' in answers:
-        settings = answers['installation-to-overwrite'].readSettings()
-        return (settings['master'] is not None)
+        try:
+            settings = answers['installation-to-overwrite'].readSettings()
+            return (settings['master'] is not None)
+        except product.SettingsNotAvailable as text:
+            logger.log("Settings unavailable: %s" % text)
     return False
 
 is_using_remote_media_fn = lambda a: 'source-media' in a and a['source-media'] in ['url', 'nfs']
@@ -72,20 +75,25 @@ def runMainSequence(results, ram_warning, vt_warning, suppress_extra_cd_dialog):
     def preserve_timezone(answers):
         if not_preserve_settings(answers):
             return False
-        if 'installation-to-overwrite' not in answers:
-            return False
-        settings = answers['installation-to-overwrite'].readSettings()
-        return 'timezone' in settings and 'request-timezone' not in settings
+        if 'installation-to-overwrite' in answers and \
+                answers['installation-to-overwrite'].settingsAvailable():
+            settings = answers['installation-to-overwrite'].readSettings()
+            return 'timezone' in settings and 'request-timezone' not in settings
+        return False
     not_preserve_timezone = lambda a: not preserve_timezone(a)
 
     def ha_enabled(answers):
         settings = {}
-        if 'installation-to-overwrite' in answers:
+        if 'installation-to-overwrite' in answers and \
+                answers['installation-to-overwrite'].settingsAvailable():
             settings = answers['installation-to-overwrite'].readSettings()
         return 'ha-armed' in settings and settings['ha-armed']
 
     def out_of_order_pool_upgrade_fn(answers):
-        if 'installation-to-overwrite' not in answers:
+        if (
+            "installation-to-overwrite" not in answers
+            or not answers["installation-to-overwrite"].settingsAvailable()
+        ):
             return False
 
         ret = False


bernhardkaindl (approved):

product.py
```diff
@@ -280,6 +280,8 @@ def fetchIfaceInfoFromNetworkdbAsDict(bridge, iface=None):
                     results['net-admin-configuration'].addIPv6(NetInterface.DHCP)
                 elif protov6 == 'autoconf':
                     results['net-admin-configuration'].addIPv6(NetInterface.Autoconf)
+            else:
+                raise SettingsNotAvailable("Could not find networkd.db on the file system to determine net-admin-bridge or net-admin-interface")
```
rosslagerwall, line 284: As these exceptions may be presented to the user in some situations, I would suggest keeping the text higher level - e.g. `Could not find network configuration`  - and then logging a message with the details.

rosslagerwall: I think you need to raise an exception in the `if not mgmt_iface:` clause as well.

rosslagerwall (commented):
Have you checked whether the callers handle the exception properly?

E.g. `setup_runtime_networking()` just ignores any exception completely (though I think in that case it might be OK because if the information can't be retrieved it will ask the user for it interactively).

product.py
GeraldEV: I think I've appropriately caught and handled the exception everywhere (it gets quite nested quickly)
Some places are for checking if we need to update the networking which is skipped if we don't have the information to update in the settings
Updated the exceptions to be user friendly and added logging

rosslagerwall (dismissed):
Can you squash fixup commits before merging? Thanks

product.py
```diff
@@ -280,6 +281,9 @@ def fetchIfaceInfoFromNetworkdbAsDict(bridge, iface=None):
                     results['net-admin-configuration'].addIPv6(NetInterface.DHCP)
                 elif protov6 == 'autoconf':
                     results['net-admin-configuration'].addIPv6(NetInterface.Autoconf)
+            else:
+                logger.log("Failed to find netword.db on filesystem")
```
bernhardkaindl, line 285: I see a typo here, what about:
```suggestion
                logger.log("Failed to find " + self.join_state_path(constants.NETWORK_DB))
```

tui/installer/__init__.py
```diff
@@ -72,20 +75,20 @@ def preserve_settings(answers):
...
+        if 'installation-to-overwrite' in answers and answers['installation-to-overwrite'].settingsAvailable():
+            settings = answers['installation-to-overwrite'].readSettings()
+            return 'timezone' in settings and 'request-timezone' not in settings
+        return False
     not_preserve_timezone = lambda a: not preserve_timezone(a)
 
     def ha_enabled(answers):
         settings = {}
-        if 'installation-to-overwrite' in answers:
+        if 'installation-to-overwrite' in answers and answers['installation-to-overwrite'].settingsAvailable():
```
bernhardkaindl, line 86: ditto
```suggestion
        if (
            "installation-to-overwrite" in answers
            and answers["installation-to-overwrite"].settingsAvailable()
        ):
```

tui/installer/__init__.py
```diff
@@ -72,20 +75,20 @@ def preserve_settings(answers):
...
             settings = answers['installation-to-overwrite'].readSettings()
         return 'ha-armed' in settings and settings['ha-armed']
 
     def out_of_order_pool_upgrade_fn(answers):
-        if 'installation-to-overwrite' not in answers:
+        if not ('installation-to-overwrite' in answers and answers['installation-to-overwrite'].settingsAvailable()):
```
bernhardkaindl, line 91: ditto / likewise, and
the `key not in dict` pattern and together with `or not` is also easier or less difficult to read for me:
```suggestion
        if (
            "installation-to-overwrite" not in answers
            or not answers["installation-to-overwrite"].settingsAvailable()
        ):
```

bernhardkaindl (changes requested):
@GeraldEV Would these suggestions good for you?

If these are not OK, please notify me.

Of course, I could revert the status to Approved otherwise.

The typo in the log message would be good to fix anyway.

tui/installer/__init__.py
GeraldEV: I've opted for the `and \` convention in keeping with the rest of this code base
Same for other comments

tui/installer/__init__.py
```diff
@@ -72,20 +75,23 @@ def preserve_settings(answers):
...
-        if 'installation-to-overwrite' in answers:
+        if 'installation-to-overwrite' in answers and \
+                answers['installation-to-overwrite'].settingsAvailable():
             settings = answers['installation-to-overwrite'].readSettings()
         return 'ha-armed' in settings and settings['ha-armed']
 
     def out_of_order_pool_upgrade_fn(answers):
-        if 'installation-to-overwrite' not in answers:
+        if 'installation-to-overwrite' not in answers and \
+                not answers['installation-to-overwrite'].settingsAvailable()):
```
bernhardkaindl, line 94: When the `not` is in the individual conditions (language-like), the `and` here needs to be an `or`.

With the `and`, the 2nd condition would be attempted when `'installation-to-overwrite' not in answers` ;-)

I'd also like to mention or discuss formatting in general:

I while ago, I was actively contributing as one of the maintainers of `https://spack.io/` (now I'm just a maintainer on leave, one could say), I was initially not so happy with the team's decision to use the black formatter for all code (and it's a huge Python code base, the complete code was re-formatted using black).

I learnt that while Black being an opinionated code formatter (having no options besides the line length and if string quoting shall be normalized to `"`-strings), it's formatting is actually very readable.

The good thing about black's formatting is, that there the formatting style is clear, there is no ambiguity how black formats it, it's always consistent.

Which is also good in another way: When using black, black formats the code, and the only variable is the maximum line length to use for black formatting.

This allows to not even have to thinking or weighting on how the code should be formatted, as there is one way it will look like after black formatted it.

Now, Of course, we can't reformat all code at once using black, as most larger patch sets would have conflicts at that point, but I'd be in favour of comitting new changes using the formatting from the tool `darker`, which formats only the added lines using black, and using it since many months, I found not even a single problem doing so.

I know, hat the formatting which black/darker provide for multi-line conditions looks as first foreign. But after some time, I've got used to it:

At least to me, it makes the such multi-line if condition more clearer to read, as the
```py
    if (
```
and the closing 
```py
    ):
```
give room to better see what's in the if condition, and the indentation of the two lines is not indented by 8 characters, they are on the same indentation level, which is good:
```suggestion
        if (
            "installation-to-overwrite" not in answers
            or not answers["installation-to-overwrite"].settingsAvailable()
        ):
```
Even if that causes two more lines of code, I think they are worth it to catch such errors better.

bernhardkaindl (changes requested):

bernhardkaindl (approved):

rosslagerwall (dismissed):

rosslagerwall (approved):

bernhardkaindl (approved):

PR #59, opened 2023-08-02
CP-20921: Support Basic HTTP Auth for answerfiles and postinstall scripts

GeraldEV:
Support using basic authentication for HTTP urls for answerfiles and install scripts for unattended installs
Tested installing from remote answerfile and postinstall script to confirm installation completed successfully, confirmed files could not be accessed without Basic Auth


util.py
```diff
@@ -250,8 +251,22 @@ def fetchFile(source, dest):
                source[:6] == 'https:' or \
                source[:5] == 'file:' or \
                source[:4] == 'ftp:':
-            # This something that can be fetched using urllib2:
-            fd = urllib2.urlopen(source)
+            # This something that can be fetched using urllib2
+            request = source
+
+            if source[:5] == 'http:' and '@' in source: # Auth HTTP
+                pattern = "http://(.+:.+)@(.+)"
```
alexbrett, line 258: Can we use urlparse here to extract the username and password rather than a regex? (https://docs.python.org/2/library/urlparse.html?highlight=urlparse#urlparse.urlparse)

bernhardkaindl: Here is the constructor from python-libs: xcp.accessor.HTTPAccessor() which is tested with Python2 and Python3:
https://github.com/xenserver/python-libs/blob/master/xcp/accessor.py#L366C1-L383C55

This test is unit-tested with Python2 and Python3 and should support also urlencoded Unicode strings for username and password (e.g. "Tanım" which is encoded as "Tan%u0131m") due to the use of urllib.parse.unquote() to get the unquoted string.

The urllib parser works with HTTPS and is unit-tested with HTTP basic auth using a simple ASCII username and password, so that's a good reminder to improve the test to use urlencoded username and password instead.

If the answerfile is red as Python2 string (bytes), and urllib just assumes UTF-8 bytes, not using unquote might work, but it should be better to use urllib.parse.unquote() to support such urlencoded username/password just in case it helps with the whole jungle of different charset encodings.

bernhardkaindl: The PR to test HTTP Basic Auth with user and password using URL-encoded Unicode is submitted:
https://github.com/xenserver/python-libs/pull/117

GeraldEV: I'll reuse the URL class defined in this file, it does URL parsing and extracts the username/password for me :)

util.py
```diff
@@ -424,3 +433,4 @@ def getPlainURL(self):
             assert self.password is None
 
             return self.url
+
```
andyhhp, line 436: Spurious addition

util.py
```diff
@@ -250,8 +251,16 @@ def fetchFile(source, dest):
...
-            # This something that can be fetched using urllib2:
-            fd = urllib2.urlopen(source)
+            # This something that can be fetched using urllib2
+            request = source
+
+            if source.startswith('http') and '@' in source: # Auth HTTP(s)
+                url = URL(source)
+                request = urllib2.Request(url.getPlainURL())
+                b64auth = base64.standard_b64encode("%s:%s" % (url.username, url.password))
+                request.add_header("Authorization", "Basic %s" % b64auth)
```
rosslagerwall, line 261: urllib2 has support for HTTP basic auth. Its IMO a silly interface but probably still better than hand-rolling it. See URLAccessor in repository.py for an example.

util.py
```diff
@@ -250,8 +251,16 @@ def fetchFile(source, dest):
                source[:6] == 'https:' or \
                source[:5] == 'file:' or \
                source[:4] == 'ftp:':
-            # This something that can be fetched using urllib2:
-            fd = urllib2.urlopen(source)
+            # This something that can be fetched using urllib2
+            request = source
+
+            if source.startswith('http') and '@' in source: # Auth HTTP(s)
```
rosslagerwall, line 257: `@` is a valid symbol to have in a URL path so this check is overzealous. Rather construct the URL object first and then use `getScheme()`/`getUsername()`/`getPassword()` to check if HTTP basic authentication needs to be set up.

rosslagerwall (approved):

ashwin9390 (approved):
Should we have an exception when fetching URL  and copyfileobject fails  ?

GeraldEV:
In the event parsing the URL, authentication or copying the file fails, it will produce a reasonable exception from the lib code already, I don't think a custom error message would help any further in handling the issue

PR #69, opened 2023-10-09
CP-44556: Support xs9

liulinC:
To support XS9, there are some updates:
* Update sfdisk command line
* Update yum to dnf
* Setup a new rpm group for test purpose
* Remove history_record and default to true The new rpm group may be revertted when we got all package ready

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1143,20 +1143,6 @@ def installBootLoader(mounts, disk, boot_partnum, primary_partnum, target_boot_m
             else:
                 installGrub2(mounts, root_partition, True)
 
-        if serial:
-            # ensure a getty will run on the serial console
-            old = open("%s/etc/inittab" % mounts['root'], 'r')
-            new = open('/tmp/inittab', 'w')
-
-            for line in old:
-                if line.startswith("s%d:" % serial.id):
-                    new.write(re.sub(r'getty \S+ \S+', "getty %s %s" % (serial.dev, serial.baud), line))
-                else:
-                    new.write(line)
-
-            old.close()
-            new.close()
-            shutil.move('/tmp/inittab', "%s/etc/inittab" % mounts['root'])
     finally:
         # done installing - undo our extra mounts:
         util.umount("%s/proc" % mounts['root'])
--- a/disktools.py
+++ b/disktools.py
@@ -826,7 +826,8 @@ class DOSPartitionTool(PartitionToolBase):
 
     def commitActivePartitiontoDisk(self, part_num):
         self.settleUdev()
-        self.cmdWrap([self.SFDISK, '--no-reread', '-A%d' % part_num, self.device]) # BIOS bootable flag set for one and unset for others partition
+        # BIOS bootable flag set for one and unset for others partition
+        self.cmdWrap([self.SFDISK, '--no-reread', '-A', self.device, part_num])
         self.waitForDeviceNodes()
 
     def writeThisPartitionTable(self, table, dryrun=False, log=False):
@@ -1045,7 +1046,7 @@ class GPTPartitionTool(PartitionToolBase):
             # at location 0x1fe.
             # However, let's keep them happy by making the single partition in the protective MBR "active".
             self.settleUdev()
-            self.cmdWrap(['sfdisk', '--no-reread', '-A1', self.device])
+            self.cmdWrap(['sfdisk', '--no-reread', '-A', self.device, '1'])
 
         # Ensure that we write out in on-disk order to prevent conflicts when
         # partition sizes get rounded.
--- a/repository.py
+++ b/repository.py
@@ -233,7 +233,7 @@ class MainYumRepository(YumRepositoryWithInfo):
     """Represents a Yum repository containing the main XenServer installation."""
 
     INFO_FILENAME = ".treeinfo"
-    _targets = ['@xenserver_base', '@xenserver_dom0']
+    _targets = ['@dom0']
 
     def __init__(self, accessor):
         super(MainYumRepository, self).__init__(accessor)
--- a/repository.py
+++ b/repository.py
@@ -782,12 +782,11 @@ def findRepositoriesOnMedia(drivers=False):
 def installFromYum(targets, mounts, progress_callback, cachedir):
         # Use a temporary file to avoid deadlocking
         stderr = tempfile.TemporaryFile()
-
-        yum_command = ['yum', '-c', '/root/yum.conf',
+        dnf_cmd = ['dnf', '--releasever=/', '-c', '/root/yum.conf',
                        '--installroot', mounts['root'],
                        'install', '-y'] + targets
-        logger.log("Running yum: %s" % ' '.join(yum_command))
-        p = subprocess.Popen(yum_command, stdout=subprocess.PIPE, stderr=stderr, universal_newlines=True)
+        logger.log("Running : %s" % ' '.join(dnf_cmd))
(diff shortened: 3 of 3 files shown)


GeraldEV:
Hi @liulinC I may have misunderstood your slack message, there's a new branch in the host installer for XS8 support (release/xs8), and the master branch will be used for XS9, so PRs to master won't need switches on the version as it will always be XS9+

backend.py
```diff
@@ -1143,7 +1143,7 @@ def installBootLoader(mounts, disk, boot_partnum, primary_partnum, target_boot_m
             else:
                 installGrub2(mounts, root_partition, True)
 
-        if serial:
+        if serial and util.getMajorVersion() < 9: # xs9 does not have inittab anymore
```
GeraldEV, line 1146: XS9 on master so this whole block can be deleted

disktools.py
```diff
@@ -826,7 +826,13 @@ def partitionTable(self):
 
     def commitActivePartitiontoDisk(self, part_num):
         self.settleUdev()
-        self.cmdWrap([self.SFDISK, '--no-reread', '-A%d' % part_num, self.device]) # BIOS bootable flag set for one and unset for others partition
+        cmd = [self.SFDISK, '--no-reread']
```
GeraldEV, line 829: Can be updated to just the XS9 syntax

disktools.py
```diff
@@ -1045,7 +1051,12 @@ def writeThisPartitionTable(self, table, dryrun=False, log=False):
             # at location 0x1fe.
             # However, let's keep them happy by making the single partition in the protective MBR "active".
             self.settleUdev()
-            self.cmdWrap(['sfdisk', '--no-reread', '-A1', self.device])
+            cmd = ['sfdisk', '--no-reread']
```
GeraldEV, line 1054: Can be updated to only XS9 syntax

repository.py
```diff
@@ -233,7 +232,7 @@ class MainYumRepository(YumRepositoryWithInfo):
     """Represents a Yum repository containing the main XenServer installation."""
 
     INFO_FILENAME = ".treeinfo"
-    _targets = ['@xenserver_base', '@xenserver_dom0']
+    _targets = ['@xenserver_base', '@xenserver_dom0'] if util.getMajorVersion() < 9 else ['@dom0']
```
GeraldEV, line 235: Just use ['@dom0'] directly

util.py
```diff
@@ -373,6 +374,9 @@ def setLocalTime(timestring, timezone=None):
 
     assert runCmd2("date --set='%s'" % timestring) == 0
 
+def getMajorVersion():
```
GeraldEV, line 377: Since master is XS9 then the usages of this will disappear so this can be removed too

GeraldEV (changes requested):

liulinC:
> Hi @liulinC I may have misunderstood your slack message, there's a new branch in the host installer for XS8 support (release/xs8), and the master branch will be used for XS9, so PRs to master won't need switches on the version as it will always be XS9+

I saw a request from alex to make master compatible with xs8 and xs9, 
so when we fix bugs, we just fix master, other than fix both master and xs8.
That is the reason why I try to make master compatible with xs8.

But if it is already agreed to branch out xs8 for this repo, I will follow.

rosslagerwall:
> I saw a request from alex to make master compatible with xs8 and xs9,
so when we fix bugs, we just fix master, other than fix both master and xs8.
That is the reason why I try to make master compatible with xs8.

Indeed that is generally preferable but in this case the large amount of changes caused by the Python 3 transition means that it is impractical to maintain a branch that works with both XS8 and XS9 so the repo was already branched in preparation for the Python 3 chanegs.

MarkSymsCtx:
> > I saw a request from alex to make master compatible with xs8 and xs9,
> > so when we fix bugs, we just fix master, other than fix both master and xs8.
> > That is the reason why I try to make master compatible with xs8.
> 
> Indeed that is generally preferable but in this case the large amount of changes caused by the Python 3 transition means that it is impractical to maintain a branch that works with both XS8 and XS9 so the repo was already branched in preparation for the Python 3 chanegs.

But, XenServer 8 has python3 available, and given that we will issue an updated ISO for XenServer 8 GA and most likely before the LTS having the hostinstaller using python3 would be a good thing and I think what Alex was asking for.

GeraldEV:
After some discussion the strategy is as follows:
master is used for XS9 and will by Python3 compatible
release/xs8 will be used for XS8 and will remain Python2 until there is suitable need/desire for it to be upgraded (which should be relatively simple after updating the dependencies)
This resolves a couple of the issues around tags being slightly out of order

rosslagerwall:
> > > I saw a request from alex to make master compatible with xs8 and xs9,
> > > so when we fix bugs, we just fix master, other than fix both master and xs8.
> > > That is the reason why I try to make master compatible with xs8.
> > 
> > 
> > Indeed that is generally preferable but in this case the large amount of changes caused by the Python 3 transition means that it is impractical to maintain a branch that works with both XS8 and XS9 so the repo was already branched in preparation for the Python 3 chanegs.
> 
> But, XenServer 8 has python3 available, and given that we will issue an updated ISO for XenServer 8 GA and most likely before the LTS having the hostinstaller using python3 would be a good thing and I think what Alex was asking for.

Changing to Python 3 in XS8 introduces a fair amount of risk which I think is undesirable _after_ a public preview, especially given the number of code paths not regularly tested. The benefit would be less maintainenance effort backporting fixes to branches but this would be counteracted by spending more effort making code compatible with both XS8 and XS9.

andyhhp:
I agree with @rosslagerwall.  Making things Py2+Py3 compatible is the same amount of effort as making things Py3-only.

But the risk of swapping it from Py2-only to Py3-only in XS8 is massive.  It's far less risk to have a single Py2+Py3 codebase that still operates in Py2 in XS8 and Py3 in XS9.

liulinC:
@rosslagerwall @GeraldEV
 Please take a look again, just updated basing on the suggestions.

ydirson:
> I agree with @rosslagerwall. Making things Py2+Py3 compatible is the same amount of effort as making things Py3-only.
> 
> But the risk of swapping it from Py2-only to Py3-only in XS8 is massive. It's far less risk to have a single Py2+Py3 codebase that still operates in Py2 in XS8 and Py3 in XS9.

Wouldn't that mean getting back to the original approach of this PR, keeping support for xs8?

GeraldEV (approved):

backend.py
```diff
@@ -1143,20 +1143,6 @@ def installBootLoader(mounts, disk, boot_partnum, primary_partnum, target_boot_m
             else:
                 installGrub2(mounts, root_partition, True)
 
-        if serial:
-            # ensure a getty will run on the serial console
```
rosslagerwall, line 1147: Doesn't this need to be reimplemented somehow instead of just dropping it?

liulinC: > Doesn't this need to be reimplemented somehow instead of just dropping it?

in xs8, we have following, So I think we can just drop it.

# inittab is no longer used when using systemd.
#
# ADDING CONFIGURATION HERE WILL HAVE NO EFFECT ON YOUR SYSTEM.
#
# Ctrl-Alt-Delete is handled by /usr/lib/systemd/system/ctrl-alt-del.target
#
# systemd uses 'targets' instead of runlevels. By default, there are two main targets:
#
# multi-user.target: analogous to runlevel 3
# graphical.target: analogous to runlevel 5
#
# To view current default target, run:
# systemctl get-default
#
# To set a default target, run:
# systemctl set-default TARGET.target
#

Issue #66, opened 2023-10-04
10.10.9 aborts in presence of a backup whose rootfs is on a "missing" disk


ydirson:
My recently-merged "tui: show the disk on which an existing installation or backup is found" patch reveals an old issue.

In a nested BIOS install a backup detected as `<XenServerBackup: XCP-ng 8.3 (8.3.0-cloud) on /dev/sda2>`, where the `PRIMARY_DISK` is `/dev/disk/by-id/nvme-QEMU_NVMe_Ctrl_nvme0` (denoting that backup was originally installed as UEFI), is presented to the user as a valid choice to restore from.

However, since 10.10.9 we now display more detailed entries, and this causes a exception, as:
- `getDiskDeviceSize("/dev/disk/by-id/nvme-QEMU_NVMe_Ctrl_nvme0")` returns `None` because of a missing `else:` branch (`dev = "disk!by-id!nvme-QEMU_NVMe_Ctrl_nvme0"` founds nothing matching `/sys/block/%s/device/block/size` or `/sys/block/%s/size`)
- `getExtendedDiskInfo()` attempts to divide `None / 2048`, which causes the installer to abort

No such problem when booting the installer from UEFI, the backup entry succeeds in getting formatted.

The `disk!by-id!nvme-QEMU_NVMe_Ctrl_nvme0` string, where in UEFI boot `/sys/block/` contains `nvme0n1` makes me suspecting we're trying to find a non-existent device, do not catch the failure to get its realpath, and go on using the bad device path.

This issue is linked to #11, in that a UEFI backup should be considered invalid on a BIOS install, and not proposed.
Note that in theory any `PRIMARY_DISK` pointing to a "missing disk" would lead to this, this could e.g. also happen if in a future platform udev rules change the `/disk/by-id/` link.

PR #70, opened 2023-10-11
Don't fail seeing a backup refering to non-existent disk

ydirson:
This fixes the various problems seen as part of https://github.com/xenserver/host-installer/issues/66, and then some.

This is a continuation of #67 which got closed when `xs8` was destroyed and cannot be reopened (thx github).

### DIFF
--- a/product.py
+++ b/product.py
@@ -604,17 +604,29 @@ def findXenSourceBackups():
         b = None
         try:
             b = util.TempMount(p, 'backup-', ['ro'])
-            if os.path.exists(os.path.join(b.mount_point, '.xen-backup-partition')):
-                backup = XenServerBackup(p, b.mount_point)
-                logger.log("Found a backup: %s" % (repr(backup),))
-                if backup.version >= XENSERVER_MIN_VERSION and \
-                        backup.version <= THIS_PLATFORM_VERSION:
-                    backups.append(backup)
+            if not os.path.exists(os.path.join(b.mount_point, '.xen-backup-partition')):
+                continue
+
+            backup = XenServerBackup(p, b.mount_point)
+            logger.log("Found a backup: %s" % (repr(backup),))
+
+            if backup.version < XENSERVER_MIN_VERSION:
+                logger.log("findXenSourceBackups: ignoring, platform too old: %s < %s" %
+                           (backup.version, XENSERVER_MIN_VERSION))
+                continue
+            if backup.version > THIS_PLATFORM_VERSION:
+                logger.log("findXenSourceBackups: ignoring later platform: %s > %s" %
+                           (backup.version, THIS_PLATFORM_VERSION))
+                continue
+
+            backups.append(backup)
+
         except Exception as ex:
             logger.info("findXenSourceBackups caught exception for partition %s", p, exc_info=1)
             pass
-        if b:
-            b.unmount()
+        finally:
+            if b:
+                b.unmount()
 
     return backups
 
--- a/product.py
+++ b/product.py
@@ -610,7 +610,8 @@ def findXenSourceBackups():
                 if backup.version >= XENSERVER_MIN_VERSION and \
                         backup.version <= THIS_PLATFORM_VERSION:
                     backups.append(backup)
-        except:
+        except Exception as ex:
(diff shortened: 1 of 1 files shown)


diskutil.py
```diff
@@ -280,6 +280,9 @@ def getDiskDeviceSize(dev):
         return int(__readOneLineFile__("/sys/block/%s/device/block/size" % dev))
     elif os.path.exists("/sys/block/%s/size" % dev):
         return int(__readOneLineFile__("/sys/block/%s/size" % dev))
+    else:
+        raise Exception("%s not found as %s or %s" % (dev, "/sys/block/%s/device/block/size",
```
alexbrett, line 284: I think this will end up raising as e.g. `sda not found as /sys/block/%s/device/block/size or /sys/block/%s/size`, i.e. the `%s` characters will come out in the exception text, which doesn't seem quite what you'd expect?

If we don't need this for xs8 (i.e. we can rely on py3 functionality), could we use a format string to make it easier, at which point you'd just have e.g. `raise Exception(f"{dev} not found as /sys/block/{dev}/device/block/size or /sys/block/{dev}/size")`?

product.py
```diff
@@ -534,13 +534,31 @@ def findXenSourceBackups():
...
+                logger.error("findXenSourceBackups: PRIMARY_DISK=%r does not exist" %
+                             (backup.root_disk,))
+                raise StopIteration()
+
+            backups.append(backup)
+
+        except StopIteration:
+            pass
+        except Exception as ex:
+            logger.log("findXenSourceBackups caught exception for partition %s: %s" % (p, ex))
```
alexbrett, line 561: We probably want to use `logger.logException(ex)` here rather than trying to include the exception directly in the log message, as that will then give us a useful traceback etc, so e.g.:
```
logger.log("findXenSourceBackups caught exception for partition %s:" % (p,))
logger.logException(ex)
```

diskutil.py
freddy77: We are trying to keep Python 2/3 compatibility, or at least have `master` and `release/xs8` branches as similar as possible. Maybe a more generic and easier `"Unable to get block size of %s device" % dev` would be enough, also not requiring to be updated as new potential virtual files are needed to be checked.

freddy77:
BTW... got the error too this morning. Pretty boring indeed.

diskutil.py
ydirson: moving the full paths inside the format string should be good for now.  At least using `str.format` and `{0}` will allow to avoid repeating `dev`

product.py
ydirson: fixed

product.py
```diff
@@ -534,13 +534,32 @@ def findXenSourceBackups():
...
             b = util.TempMount(p, 'backup-', ['ro'], 'ext3')
-            if os.path.exists(os.path.join(b.mount_point, '.xen-backup-partition')):
-                backup = XenServerBackup(p, b.mount_point)
-                logger.log("Found a backup: %s" % (repr(backup),))
-                if backup.version >= XENSERVER_MIN_VERSION and \
-                        backup.version <= THIS_PLATFORM_VERSION:
-                    backups.append(backup)
-        except:
+            if not os.path.exists(os.path.join(b.mount_point, '.xen-backup-partition')):
+                raise StopIteration()
```
freddy77, line 538: I found this raising `StopIteration` and catching it weird and hard to read. Why not using a `try/finally` doing the `b.umount()` in the finally and using `continue` here?

ydirson: The lack of `break` equivalent for `try` blocks is a problem.  I guess `return []` here would be suitable enough.

ydirson:
Rebased, and improved the exception logging in patch 2

product.py
freddy77: You are inside a `for` loop, why not using `break` and/or `continue` ? Also, you can put the `b.unmount()` inside a `finally` section.

ydirson: good point :sweat_smile:

ydirson: fixed

diskutil.py
ydirson: This code has be replaced

ydirson:
Rebased, no conflicts.  Is there anything holding back this PR?

PR #71, opened 2023-10-11
installFromYum: give more detailed error messages on gpg errors

ydirson:
This is a complement to the gpg-checking feature enabled for 8.3.

This is a continuation of #68 which got closed when xs8 was destroyed and cannot be reopened (thx github).

Covers:
1. repo_gpgcheck:
  a. wrong system clock putting gpg key creation in the future, causing a yum crash (nothing special happens if the date of the signature is in the future `¯\_(ツ)_/¯`)
  b. other yum crashes due to uncaught gpg exceptions (if any)
  c. lack of repomd signature (while repo_gpgcheck is in force)
  d. signature done by other key than the one in ISO ("repomd.xml signature could not be verified" `¯\_(ツ)_/¯`)
3. gpgcheck:
  a. RPM signed with unknown key
  b. unsigned RPM referenced by unsigned repomd (no-repo-gpgcheck)
  c. RPM re-signed with unknown key, unsigned repomd (no-repo-gpgcheck)
  d. RPM overwritten with another RPM signed with known key (diagnosed through hash but, same diag as 2.c)
  e. delsigned/resigned/etc RPM, unchanged repomd (same diag as 2.c/d)

Does not cover notably:
  - unsigned RPM referenced by (re)signed repomd

In some cases Yum does not give an error, but dies because of an uncaught exception, which makes this check quite brittle, but in the worst case if messages change, we still fallback to the original "Error installing packages" message.

### DIFF
--- a/repository.py
+++ b/repository.py
@@ -844,10 +844,38 @@ def installFromYum(targets, mounts, progress_callback, cachedir):
                 total = 0
                 updateInstallProgress(m)
         rv = p.wait()
+        gpg_error_message = None
+
+            if stderr.find(' in import_key_to_pubring') >= 0:
+                gpg_error_message = "Signature key import failed"
+            # add any other instance of uncaught GpgmeError before this like
+            elif stderr.find('gpgme.GpgmeError: ') >= 0:
+                gpg_error_message = "Cryptography-related yum crash"
+
+            elif re.search("Couldn't open file [^ ]*/repodata/repomd.xml.asc", stderr):
+                # would otherwise be mistaken for "pubring import" !?
+                gpg_error_message = "No signature on repository metadata"
+            elif stderr.find('repomd.xml signature could not be verified') >= 0:
+                gpg_error_message = "Repository signature verification failure"
+
+            else:
+                match = re.search("Public key for ([^ ]*.rpm) is not installed", stderr)
+                if match:
+                    gpg_error_message = "Missing key for %s" % (match.group(1),)
+                match = re.search("Package ([^ ]*.rpm) is not signed", stderr)
+                if match:
+                    gpg_error_message = "Package not signed: %s" % (match.group(1),)
+                match = re.search(r" ([^ ]*): \[Errno [0-9]*\] No more mirrors to try", stderr)
+                if match:
+                    # rpm not found or corrupted/re-signed/etc
+                    gpg_error_rpm_not_found = match.group(1)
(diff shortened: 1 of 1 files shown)


repository.py
```diff
@@ -821,12 +821,59 @@ def installFromYum(targets, mounts, progress_callback, cachedir):
         rv = p.wait()
         stderr.seek(0)
         stderr = stderr.read()
+        gpg_uncaught_error = 0
```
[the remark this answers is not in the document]
ydirson: I cannot recall precisely, but:
- (likely original reason) to keep the structure of the existing code; notably, there was already an exception raised with a generic message, and all that needed to be changed was the message
- in line with that, it allows to keep the "yum exited" log, which would have to be duplicated, or wrapped with the exception in a helper (and delegating the raise to a helper function causes its own problems)
- additionally it allows to make sure to use the most pertinent message, regardless of the matching order, which in turn is constrained by the order in which strings appear.  Maybe not really necessary for this set of messages, but considering this is ad-hoc parsing of a program's output, including parsing of its crashes, it may be useful to retain this flexibility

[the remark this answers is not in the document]
freddy77: Why instead of using a lot of flags variable you don't just set a string variable `errmsg` to save the specific error and use it? You could initialize it at default generic message and change accordingly based on the specific error.

ydirson:
Last push rebases onto latest xs8

repository.py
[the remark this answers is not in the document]
ydirson: Applied suggestion

ydirson:
Now based on `master`, with review comment taken into account

repository.py
```diff
@@ -844,10 +844,38 @@ def updateInstallProgress(m):
                 total = 0
                 updateInstallProgress(m)
         rv = p.wait()
+        gpg_error_message = None
+
+            if stderr.find(' in import_key_to_pubring') >= 0:
```
freddy77, line 849: Why this extra indentation?

ydirson: It's not really easy for me to do such forward-porting on `master` (especially testing), since we've not started to use that branch.  Would it be possible to target xs8 first, and then once that gets accepted start to work on a `master` version, to streamline the process a bit more?

freddy77: Sure, fine with me

ydirson: It took some time to get back to it, but #334 is here at last.

PR #73, opened 2023-10-18
Improve Driver Disks handling as Supplemental Packs

ydirson:
This PR improves the final "Supplemental Packs" dialog when the user made use of Driver Disk(s), and also fixes the handling of Driver Disks when the supplemental-packs feature is disabled (re-enabling the dialog, with the more targeted message).


tui/installer/screens.py
```diff
@@ -482,10 +485,14 @@ def confirm_erase_volume_groups(answers):
     return RIGHT_FORWARDS
 
 def use_extra_media(answers):
+    message = "You installed Would you like to install any Supplemental Packs?"
```
alexbrett, line 488: This doesn't seem to read correctly - should the `You installed` be there?

tui/installer/screens.py
```diff
@@ -482,10 +485,14 @@ def confirm_erase_volume_groups(answers):
...
+    if "driver-repos" in answers:
+        message += ("  You loaded one or more Driver Disks during this installation, "
+                    "you likely want to install them now using this feature.")
```
alexbrett, line 491: Would this be better as "You previously loaded one or more Driver Disks, if you wish to include these drivers on the installed system, you must install them now"?

tui/installer/screens.py
ydirson: Right, changed my mind change in the middle of it and missed that.

tui/installer/screens.py
ydirson: Fine with me, will reroll.

stormi:
I did check they were resolved, at least to my understanding :). I rarely ping despite many pending reviews, by the way.

freddy77 (dismissed):

stormi:
If you meant the conversations were still open from github's point of view, it's not always obvious in review processes whose responsibility it is to mark them as resolved. The committer, or the reviewer? I usually prefer when it's the reviewer, this way they have a way to have a last look to what they asked and how it was answered, before marking the conversation as resolved. Others may prefer that they are marked as resolved by the committer.

freddy77:
> If you meant the conversations were still open from github's point of view, it's not always obvious in review processes whose responsibility it is to mark them as resolved. The committer, or the reviewer? I usually prefer when it's the reviewer, this way they have a way to have a last look to what they asked and how it was answered, before marking the conversation as resolved. Others may prefer that they are marked as resolved by the committer.

It sounds reasonable. Personally on my changes if I changed as requested/suggested I put a `done` comment and mark as resolved, otherwise I put a comment for the specific reviewer (why I didn't exactly follow the request or why I don't think it should be done, for instance because it will be done separately).

freddy77 (dismissed):

ydirson:
any news on this one?

freddy77:
@ydirson not much news from me, I don't understand why this got stuck.

ydirson:
Rebased on current master, let's hope we can get this PR unstuck

ydirson:
Pinging on this one as well :slightly_smiling_face:

PR #74, opened 2023-10-24
CA-384265: Remove use of cmp to fix XS9 installation

qinzhang22:
XS9 installation failed due to *cmp* is used as *cmp* is only available in python2 while XS9 is based on python3

### DIFF
--- a/tui/installer/screens.py
+++ b/tui/installer/screens.py
@@ -537,8 +537,7 @@ def disk_more_info(context):
     return True
 
 def sorted_disk_list(): # Smallest to largest, then alphabetical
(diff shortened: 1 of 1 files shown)


stormi:
The best commit message usually describe what changes, starting with a verb in the imperative form. Maybe something like "Remove use of cmp to fix XS9 installation", then an explanation in the body of the commit message stating why cmp is an issue in the context of XS9?

liulinC (approved):

andyhhp (approved):

rosslagerwall (approved):

PR #75, opened 2023-10-24
XSI-1498: Log any exceptions parsing existing bootloader config

alexbrett:
Previously any exceptions which occurred were silently suppressed, instead log such occurrences to enable easier debugging from the installer log.

### DIFF
--- a/product.py
+++ b/product.py
@@ -393,8 +393,9 @@ class ExistingInstallation:
             pciback = next((x for x in kernel_args if x.startswith('xen-pciback.hide=')), None)
             if pciback:
                 results['host-config']['xen-pciback.hide'] = pciback
-        except:
-            pass
+        except Exception as e:
+            logger.log('Exception whilst parsing existing bootloader config:')
+            logger.logException(e)
         self.unmount_boot()
 
         return results


andyhhp (approved):

ydirson (approved):

MarkSymsCtx (approved):

product.py
```diff
@@ -393,8 +393,9 @@ def fetchIfaceInfoFromNetworkdbAsDict(bridge, iface=None):
             pciback = next((x for x in kernel_args if x.startswith('xen-pciback.hide=')), None)
             if pciback:
                 results['host-config']['xen-pciback.hide'] = pciback
-        except:
-            pass
+        except Exception as e:
```
rosslagerwall, line 396: Logging an error is an improvement but should this `raise SettingsNotAvailable(...)` instead?

There are also a few `try... except ... pass` constructs further up in this function which it would be good to sort out as well (can open a ticket for it if you want).

alexbrett: I thought about raising an exception, but the problem is if something is unexpected on the old install, you probably don't want to break the ability to upgrade (better to just use 'new' defaults so at least the end user can make progress).

Interestingly a few other places that raise exceptions then get subsequently caught by an exception handler and thus are suppressed anyway (e.g. the network type detection)...

The others would probably be good to sort out, so yes I guess worth opening a ticket to improve them at the same time.

PR #77, opened 2023-11-28
CA-385350 Fix handling of NTP configuration over upgrade

alexbrett:
Previously we were not setting `ntp-config-method` when parsing the NTP configuration from an old version, resulting in any NTP servers that were read being ignored.

Set the appropriate `ntp-config-method`, and also filter out any 'default' NTP servers from the `ntp-servers` list so we only load customer specified servers. Note this is done using a fixed list of default domains, as this has changed in the past and may change in future, so cannot be read from the current release config.


alexbrett:
XenRT tests have been written to validate the behaviour, and run successfully (albeit on an xs8 version of the change, which I'll raise a separate PR for after this one is agreed)

GeraldEV (approved):

rosslagerwall (approved):

PR #86, opened 2023-12-05
CP-45554: Implement support for upgrading from old MBR to new GPT layout

rosslagerwall:
See the commit messages for details.

This is going to a feature branch for now as it needs additional QA.


freddy77:
Also https://github.com/xenserver/host-installer/pull/86/commits/5fb2094ade063e145daf30710d554c01c02563cd can go directly into `master` and `release/xs8`. Specifically last usage of `root_size_old` and `backup_size_old` were in commit e5df6f8aaa741c9d68545d0ee1d36de3f0fea71c (`CP-32753 - Remove OEM utility partition support`).

GeraldEV (approved):

freddy77:
Patch for Python 3 compatibility https://github.com/freddy77/host-installer/commit/5465e215cad1edb1449956dec9feadbf86ef77c7

constants.py
```diff
@@ -84,19 +84,22 @@ def error_string(error, logname, with_hd):
...
 logs_free_space = 20
 
+install_size = logs_size + backup_size + root_size + boot_size + swap_size
+
+# Number of sectors consumed by the GPT on a disk with 512 byte sectors.
+# 1 sector for MBR, 1 sector for GPT header + 32 sectors for partition entries.
+gpt_header_512_sectors = 34
+# Number of sectors consumed by the backup GPT on a disk with 512 byte sectors.
+# 1 sector for GPT header + 32 sectors for partition entries.
+backup_gpt_header_512_sectors = gpt_header_512_sectors - 1
```
freddy77, line 101: These should be functions depending on sector size, not assuming the sector size is always 512.

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+    lvm.file.seek(lvm.raw_locs[0].offset + start, os.SEEK_SET)
+    buf = lvm.file.read(lvm.raw_locs[0].size - wrap)
+
+    if wrap:
+        lvm.file.seek(start + MDA_HEADER_SIZE, os.SEEK_SET)
+        buf += lvm.file.read(wrap)
+
+    assert(calc_crc(buf) == lvm.raw_locs[0].checksum)
+
+    lvm.text_config = buf[:-1] # ignore NUL terminator
```
freddy77, line 361: Is it mandatory to have the `NUL` terminator? Does the standard specify some encoding for this text?

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+        l = l.strip()
+        if not l:
+            continue
+
+        l = l.rsplit('#')[0].strip()
+        if not l:
+            continue
+
+        before_state = state
+        if l == 'physical_volumes {':
```
freddy77, line 499: Is it possible to have more spaces between `physical_volumes` and `{` ? Or even a new line?

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+    pe_count = int(lvm.vgs[0].pvs[0].get_config('pe_count'))
+    start_shrunk_extents = int(math.ceil(start_space / float(extent_size)))
+    end_shrunk_extents = int(math.ceil(end_space / float(extent_size)))
+    regions = [ReservedRegion(0, start_shrunk_extents)]
+    if end_shrunk_extents > 0:
+        regions.append(ReservedRegion(pe_count - end_shrunk_extents, end_shrunk_extents))
+
+    # If we're shrinking the start by an amount that is not a multiple of the extent size
+    # increase the size of the metadata area to compensate
+    extra_metadata_sectors = (start_shrunk_extents * extent_size - start_space) / SECTOR_SIZE
```
freddy77, line 954: Just to be safe I would use `// SECTOR_SIZE`. I suppose for some weird reason LVM2 assume 512 sector size that is that is use that assumption as base unit instead of assuming the disk has always that sector size.

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+
+        return subprogress_callback
+
+    move_ops_cb = get_move_ops_progress_cb()
+    for op in move_ops:
+        move_extents(lvm, move_ops_cb, *op)
+
+    # A crash up to this point shouldn't affect anything because we've only
+    # written into free unallocated extents. Now enter the critical region and
+    # commit the changes.
```
freddy77, line 992: Potentially we could update the layout in the current partition so the next write will also update free space. But not a big issue I suppose

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+    lvm.file.write(mda_data)
+
+
+def shrink_partition(drive, partno, start_shrunk_sectors, end_shrunk_sectors):
+    '''Shrink the partition to match the new LVM size'''
+
+    tool = PartitionTool(drive)
+    old_partition = tool.partitions[partno]
+    part_start = old_partition['start']
+    part_size = old_partition['size']
```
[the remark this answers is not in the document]
freddy77: I would at least put a `assert(tool.sectorSize == SECTOR_SIZE)`

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+
+            # The original segment is removed from the allocation table and three are created:
+            # 1) A segment for data before the moved data
+            # 2) A segment for data after the moved data
+            # 3) A segment for the new location of the moved data
+            #
+            # In some cases (1) and/or (2) may contain 0 extents and so
+            # they are skipped.
+
+            # Insert (2) into the allocation table
```
freddy77, line 734: I think this should be `# Insert (3) into`...

upgrade.py
```diff
@@ -213,9 +222,102 @@ def testUpgradeForbidden(self, tool):
...
+        # Ensure new partitions do not exist.
+        boot_partition = tool.getPartition(boot_partnum)
+        assert(boot_partition is None)
+        logs_partition = tool.getPartition(logs_partnum)
+        assert(logs_partition is None)
+        swap_partition = tool.getPartition(swap_partnum)
+        assert(swap_partition is None)
+
+        # This upgrade process is only supported for 512 byte sectors.
+        assert(tool.sectorSize == 512)
```
freddy77, line 272: Good, this will make the sector size assumption work

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+    for l in lvm.text_config.split('\n'):
+        l = l.strip()
+        if not l:
+            continue
+
+        l = l.rsplit('#')[0].strip()
+        if not l:
+            continue
+
+        before_state = state
```
freddy77, line 498: This variable is never used

shrinklvm.py
freddy77: Here and below, for `logical_volumes {`, should the current state be checked? I suppose we expect these groups be be in specific state, like `IN_VG_I`.

shrinklvm.py
rosslagerwall: I haven't seen documentation that describes the details of the on-disk format. I believe the implementation always uses a NUL terminator: https://github.com/lvmteam/lvm2/blob/bb384f84880a183ee58a18935851eece25cdada4/lib/format_text/export.c#L1053

The implementation also seems to use C strings without an explicit encoding. Using Python strings is probably the most suitable data type to use.

upgrade.py
```diff
@@ -201,6 +208,8 @@ def __init__(self, source):
         self.safe2upgrade = os.path.isfile(safe2upgrade_path)
         self.vgs_output = None
 
+        self.safe2mbrupgrade = os.path.isfile(os.path.join(primary_fs.mount_point, constants.SAFE_2_MBR_UPGRADE))
```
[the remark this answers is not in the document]
rosslagerwall: A feature flag would be appropriate if choosing to enable the functionality in the host installer. However, this is a bit different. The upgrade is enabled based on the presence of the marker file in the installation itself. This may or may not be present depending on the component. It is expected that the component that initiates the update sets the marker file if needed. This follows the established convention for partition layout upgrades.

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+
+
+class _ConfigStore(object):
+    '''Store and retrieve LVM config'''
+
+    def __init__(self):
+        # The config is stored as a list of (string, string) tuples
+        # to maintain order and avoid having to parse the values
+        # which may be a variety of types.
+        self.config = []
```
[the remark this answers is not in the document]
rosslagerwall: I thought about that and then decided against it for some reason. Let me check again if it is suitable.

upgrade.py
```diff
@@ -213,9 +222,102 @@ def testUpgradeForbidden(self, tool):
...
+
+        # The new installation start is aligned up to a 1 MiB boundary.
+        # The SR will therefore be resized to a 1 MiB boundary because
+        # the installation is sized in whole MiBs.
+        new_install_start = (existing_install_start + 2047) & ~2047
+        new_sr_start = new_install_start + (constants.install_size * 1024 * 1024 / tool.sectorSize)
+        start_sectors_required = new_sr_start - tool.partitions[storage_partnum]['start']
+        end_sectors_required = max(tool.partitions[storage_partnum]['start'] +
+                                   tool.partitions[storage_partnum]['size'] -
+                                   (tool.sectorExtent - constants.backup_gpt_header_512_sectors), 0)
```
freddy77, line 291: I don't understand why we need to reserve an extent at the end of the partition. I mean, I understand there could be an extent that should be moved to make space for the GPT backup, but should not be done here, here we are computing the space for the partition, not the LVM extents.

Also the `max` here looks a bit weird; if the computation would produce a negative number it would look to be a problem, not something to silence.

shrinklvm.py
rosslagerwall: I'll fix that.

LVM2 (at least with the version we use) does seem to have a hardcoded sector size of 512: https://github.com/lvmteam/lvm2/blob/bb384f84880a183ee58a18935851eece25cdada4/lib/metadata/metadata-exported.h#L32

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+
+    # Create a sorted allocation table of the segments
+    alloc_table = lvm_to_alloc_table(lvm)
+    logger.log('Initial allocation table: {}'.format(alloc_table))
+
+    extent_size = int(lvm.vgs[0].get_config('extent_size')) * SECTOR_SIZE
+    pe_count = int(lvm.vgs[0].pvs[0].get_config('pe_count'))
+    start_shrunk_extents = int(math.ceil(start_space / float(extent_size)))
+    end_shrunk_extents = int(math.ceil(end_space / float(extent_size)))
+    regions = [ReservedRegion(0, start_shrunk_extents)]
```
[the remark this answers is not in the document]
rosslagerwall: That's not a good assumption to make... I'll fix that.

shrinklvm.py
rosslagerwall: Sorry, I'm not sure what you mean. Could you clarify please?

freddy77: More of an improvement to avoid data loss... or better make possible a recovery. Now you moved the extents to free space. If you update the LVM headers at the current position and a restart/crash happens (for whatever reason) the LVs will still have the same data. If you don't have a restart (which should be always) you will be writing the new LVM headers in the new position which are now marked as free. Now, if you have a restart/crash here the LVM will point to old headers so LVs will still have the same data. Yes, at the point of rebuilding the partition schema then you could loose data, but rewriting the LVM headers will decrease the critical region.

As said, mainly paranoia mode here...

shrinklvm.py
rosslagerwall: According to the LVM codebase, it is always a single space between the initial string and the opening brace. However, I'll update the code to be a bit more robust and accept an arbitrary amount of space (including none).

I'll also check that state is correct for physical_volumes and logical_volumes.

shrinklvm.py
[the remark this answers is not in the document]
rosslagerwall: I will switch to using OrderedDict.

shrinklvm.py
rosslagerwall: I will change this to strip any amount of NULs from the end in case it is not present.

shrinklvm.py
rosslagerwall: OK, that makes sense. I'll add it.

To reduce the window further, I'll also update the code to do the shrink partition stuff earlier and just let it do the final `commit()` at the end.

shrinklvm.py
rosslagerwall: That was some left over debugging. Removed.

shrinklvm.py
```diff
@@ -0,0 +1,1012 @@
...
+    check_lvm(lvm)
+
+    # Create a sorted allocation table of the segments
+    alloc_table = lvm_to_alloc_table(lvm)
+    logger.log('Initial allocation table: {}'.format(alloc_table))
+
+    extent_size = int(lvm.vgs[0].get_config('extent_size')) * SECTOR_SIZE
+    pe_count = int(lvm.vgs[0].pvs[0].get_config('pe_count'))
+    start_shrunk_extents = int(math.ceil(start_space / float(extent_size)))
+    end_shrunk_extents = int(math.ceil(end_space / float(extent_size)))
```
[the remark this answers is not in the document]
rosslagerwall: Will fix.

freddy77:
Tested, works also here!

freddy77 (approved):

GeraldEV (approved):

PR #90, opened 2023-12-19
CP-37929: Remove legacy logrotate mechanism

GeraldEV:
Legacy logrotate mechanism files are also being removed from the xenserver-release package.
Tested that current logrotate mechanism still works after removal and the host installer (both attended and unattended installs) work as expected.


freddy77 (approved):
Nice! More of these patches!

bernhardkaindl (approved):
Giving a 2nd approval based on freddy77's praise.

PR #92, opened 2023-12-20
Remove useless check - TRIVIAL

freddy77:
partitions has always at least one elements (a dummy None one) so if can't have a length of 0.

### DIFF
--- a/disktools.py
+++ b/disktools.py
@@ -592,9 +592,7 @@ class PartitionToolBase:
         partitions = [None] + [part for num, part in sorted(self.partitions.items(), key=lambda item: item[1]['start'])]
 
         if startBytes is None:
-            if len(partitions) == 0:
-                startSector = self.sectorFirstUsable
-            elif order:
+            if order:
                 if order < 1:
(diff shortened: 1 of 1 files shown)


disktools.py
```diff
@@ -592,9 +592,7 @@ def createPartition(self, id, sizeBytes=None, number=None, order=None, startByte
         partitions = [None] + [part for num, part in sorted(self.partitions.items(), key=lambda item: item[1]['start'])]
 
         if startBytes is None:
-            if len(partitions) == 0:
```
GeraldEV, line 595: Without this check (and with the default [None] value for partitions) won't line 603 fail if it's ever reached? Since partitions[-1] will be None and then the lookup for ["start"] and ["size"] will fail?

freddy77: That would happen with or without this test.

GeraldEV: Ah you're right, I forgot to consider the None when checking the length! :doh:

GeraldEV (approved):

rosslagerwall (approved):

PR #89, opened 2023-12-19
Minor changes to inspectTargetDisk and probeDisk

freddy77:
This series is a preparation before adding support for restoring some old systems from backup.

Changes:
- rename confusing `inspectTargetDisk` and added a comment;
- `probeDisk` returns a new object to make easier to update and add results;
- removed unused parameter from `probeDisk`;
- add `swap` result to `probeDisk`.

See commit messages for details.

PR #93, opened 2023-12-20
Allows to restore really old layout used by NetScaler SDX machines

freddy77:
This change allows to restore SDX machines updated to XenServer 8.

SDX still uses MBR and 3 partition schema (plus one possible utility partition). 
The possibility to restore this schema was removed from installer time ago but we need to add back to support restore of SDX products.

After restoring backup data to logs (which used to be the root partition) all additional partitions (EFI boot, swap, logs) not necessary are removed, backup recreated and partition schema is moved from GPT to MBR.

This PR is based on https://github.com/xenserver/host-installer/pull/89 (which contain base improves).

Opened as Draft as I don't like some aspect of it. Maybe just paranoia:

- clearing "manually" the GPT after `sfdisk`... but wiping every entirely before open a window for loosing data which is worst;
- the `Disk` object naming for the return of `probeDisk`... maybe something more like product?
- restore code looks more dirty having to support such an old schema. But not sure how to separate in another way;
- I found weird having to manually convert from GPT to MBR... why this is not done in `disktools.py`?

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -526,7 +526,7 @@ def configureNTP(mounts, ntp_config_method, ntp_servers):
 
 # This is attempting to understand the desired layout of the future partitioning
 # based on options passed and status of disk (like partition to retain).
-# This is used on upgrade or restore.
+# This should be used for upgrade or install, not for restore.
 # Returns 'target-boot-mode', 'boot-partnum', 'primary-partnum', 'backup-partnum', 'logs-partnum', 'swap-partnum', 'storage-partnum'
 def partitionTargetDisk(disk, existing, preserve_first_partition, create_sr_part):
     logger.log("Installer booted in %s mode" % ("UEFI" if constants.UEFI_INSTALLER else "legacy"))
--- a/backend.py
+++ b/backend.py
@@ -104,7 +104,7 @@ def getPrepSequence(ans, interactive):
         Task(util.getUUID, As(ans), ['installation-uuid']),
         Task(util.getUUID, As(ans), ['control-domain-uuid']),
         Task(util.randomLabelStr, As(ans), ['disk-label-suffix']),
-        Task(inspectTargetDisk, A(ans, 'primary-disk', 'installation-to-overwrite', 'preserve-first-partition','sr-on-primary'), ['target-boot-mode', 'boot-partnum', 'primary-partnum', 'backup-partnum', 'logs-partnum', 'swap-partnum', 'storage-partnum']),
+        Task(partitionTargetDisk, A(ans, 'primary-disk', 'installation-to-overwrite', 'preserve-first-partition','sr-on-primary'), ['target-boot-mode', 'boot-partnum', 'primary-partnum', 'backup-partnum', 'logs-partnum', 'swap-partnum', 'storage-partnum']),
         ]
 
     if ans['ntp-config-method'] in ("dhcp", "default", "manual"):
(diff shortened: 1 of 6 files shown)


restore.py
```diff
@@ -59,6 +59,65 @@ def restoreFromBackup(backup, progress=lambda x: ()):
...
+    if 'BOOT' not in backup_partition_layout and 'LOG' not in backup_partition_layout:
+        new_restore_partition = restore_partition
+        efi_boot = False
+
+        # remove unneeded partitions
+        if dsk.swap[0]:
+            tool.deletePartition(tool.partitionNumber(dsk.swap[1])) # swap
+        tool.deletePartitionIfPresent(boot_partnum) # boot
+
+        # use logs as root
```
rosslagerwall, line 73: It's probably worth sanity checking that the resulting partition is actually the expected size (i.e. root_mbr_size_old).

restore.py
```diff
@@ -59,6 +59,65 @@ def restoreFromBackup(backup, progress=lambda x: ()):
...
+        root_partnum = tool.partitionNumber(dsk.root[1])
+        backup_partnum = tool.partitionNumber(backup_partition)
+        if dsk.logs[0]:
```
rosslagerwall, line 76: The alignment of partitions may be fixed up during upgrade.

This doesn't restore the original (mis-)alignment during backup. I think that is fine, but it might be worth noting in a comment.

restore.py
```diff
@@ -59,6 +59,65 @@ def restoreFromBackup(backup, progress=lambda x: ()):
...
+            logs_partnum = tool.partitionNumber(dsk.logs[1])
+            if tool.partitionStart(logs_partnum) < tool.partitionStart(root_partnum) and tool.partitionSize(logs_partnum) >= (3500 << 20):
+                restore_partition = dsk.logs[1]
+                tool.deletePartition(root_partnum)
+                tool.renamePartition(logs_partnum, root_partnum)
+            else:
+                tool.deletePartition(tool.partitionNumber(dsk.logs[1])) # logs
+
+        # rewrite with MBR format, we need to manually convert IDs from GPT to MBR
```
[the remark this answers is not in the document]
freddy77: If there was no BOOT partition it couldn't have been GPT, or at least not if using grub

backend.py
```diff
@@ -524,7 +524,11 @@ def configureNTP(mounts, ntp_config_method, ntp_servers):
     util.runCmd2(['chroot', mounts['root'], 'systemctl', 'enable', 'chronyd'])
     util.runCmd2(['chroot', mounts['root'], 'systemctl', 'enable', 'chrony-wait'])
 
-def inspectTargetDisk(disk, existing, preserve_first_partition, create_sr_part):
+# This is attempting to understand the desired layout of the future partitioning
+# based on options passed and status of disk (like partition to retain).
+# This is used on upgrade or restore.
```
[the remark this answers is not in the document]
freddy77: Actually the comment is wrong, it should not be used for restore! Good catch

restore.py
freddy77: Another option would be to remove all additional partitions earlier, create new root with expected size. So another round of repartitioning. I'll think more about

restore.py
```diff
@@ -168,11 +236,30 @@ def restoreFromBackup(backup, progress=lambda x: ()):
...
-        logs_part = partitionDevice(disk, logs_partnum)
-        swap_part = partitionDevice(disk, swap_partnum)
+        logs_part = dsk.logs[1]
+        swap_part = dsk.swap[1]
         if util.runCmd2(['e2label', logs_part, constants.logsfs_label%rdm_label]) != 0:
             raise RuntimeError("Failed to label logs partition")
         if util.runCmd2(['swaplabel', '-L', constants.swap_label%rdm_label, swap_part]) != 0:
             raise RuntimeError("Failed to label swap partition")
+
+def clear_gpt_headers(tool):
```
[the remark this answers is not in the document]
freddy77: That's explained by above comment "# We do this way to avoid intermediate states without partitions".

restore.py
freddy77: Now that you mention not sure which alignment will the new backup have... surely the "old" logs will have 1MB alignment.

freddy77: Added a comment

restore.py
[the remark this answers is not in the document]
freddy77: After a discussion with Ross we decided to use MBR only if we upgraded from MBR checking a file in `/var/preserve`.
In case of GPT we need to install grub2 on the partition instead of the disk (there's an option in the answerfile, we need to update it automatically).

restore.py
freddy77: Checked size against `constants.root_gpt_size_old`

restore.py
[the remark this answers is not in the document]
freddy77: Checked and implemented but looks like `grub-install` does not do a great job restoring on the partition making the system not bootable

rosslagerwall (approved):

GeraldEV (approved):

PR #80, opened 2023-11-30
avoid duplicating contact text for error messages

ydirson:

PR #99, opened 2024-01-29
Rewrite error_string function

freddy77:
Avoids duplication of part of the error messages.
Do not declare enumeration or dictionary, just use strings.
Make sure error is terminated with a dot.

This is intended to supersedes https://github.com/xenserver/host-installer/pull/80.

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -51,25 +51,19 @@ NETWORK_BACKEND_VSWITCH_ALT = "vswitch"
 
 # error strings:
 def error_string(error, logname, with_hd):
-    (
-        ERROR_STRING_UNKNOWN_ERROR_WITH_HD,
-        ERROR_STRING_UNKNOWN_ERROR_WITHOUT_HD,
-        ERROR_STRING_KNOWN_ERROR
-    ) = list(range(3))
-
-    ERROR_STRINGS = {
-        ERROR_STRING_UNKNOWN_ERROR_WITH_HD: "An unrecoverable error has occurred.  The details of the error can be found in the log file, which has been written to /tmp/%s (and /root/%s on your hard disk if possible).\n\nPlease refer to your user guide or contact a Technical Support Representative for more details.",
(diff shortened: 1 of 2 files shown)


GeraldEV (approved):

ydirson (approved):

test/test_errors.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+import sys
+import os.path
+sys.path.append(os.path.join(os.path.abspath(os.path.dirname(__file__)), '..'))
+
+import constants
+
+def check(error, logname, with_hd, expected):
+	got = constants.error_string(error, logname, with_hd)
```
rosslagerwall, line 10: Is this using tabs?

constants.py
```diff
@@ -51,25 +51,19 @@
...
         if with_hd:
-            return ERROR_STRINGS[ERROR_STRING_UNKNOWN_ERROR_WITH_HD] % (logname, logname)
-        else:
-            return ERROR_STRINGS[ERROR_STRING_UNKNOWN_ERROR_WITHOUT_HD] % logname
+            err += " (and /root/%s on your hard disk if possible)" % logname
     else:
-        return ERROR_STRINGS[ERROR_STRING_KNOWN_ERROR] % error
+        err = "The error was:\n\n%s" % error
+
+    if err[-1:] != '.':
```
rosslagerwall, line 62: I think this doesn't need to be a slice... `err[-1]` should work.

test/test_errors.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+import sys
+import os.path
+sys.path.append(os.path.join(os.path.abspath(os.path.dirname(__file__)), '..'))
+
+import constants
+
+def check(error, logname, with_hd, expected):
```
rosslagerwall, line 9: I think it would be better to use a unit test framework for this (like `unittest`), especially as we get more unit tests over time.

constants.py
freddy77: Not if the string is empty, so slice is more "safe"

GeraldEV (approved):

constants.py
rosslagerwall: But `err` is set to a non-empty string in both parts of the if statement.

rosslagerwall (approved):

PR #101, opened 2024-02-01
CP-47538: Deprecate legacy BIOS boot

rosslagerwall:
If the warning is shown, it will be shown after agreeing to the EUA.

This is what it looks like:

deprecated

### DIFF
--- a/tui/installer/screens.py
+++ b/tui/installer/screens.py
@@ -142,7 +142,7 @@ If %s are present you may need to load a device driver on the previous screen fo
     return RIGHT_FORWARDS
 
 def hardware_warnings(answers, ram_warning, vt_warning):
-    vt_not_found_text = "Hardware virtualization assist support is not available on this system.  Either it is not present, or is disabled in the system's BIOS.  This capability is required to start Windows virtual machines."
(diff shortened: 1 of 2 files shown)


alexbrett (approved):

tui/installer/screens.py
```diff
@@ -178,6 +178,26 @@ def overwrite_warning(answers):
     if button == 'back': return LEFT_BACKWARDS
     return RIGHT_FORWARDS
 
+def bios_warning(answers):
+    text = ("Booting %s hosts in BIOS mode is now deprecated. "
+            "You can still install your %s hosts in BIOS boot mode. "
+            "However, doing so can prevent you from upgrading your %s "
```
alexbrett, line 184: Perhaps 'may prevent' rather than 'can prevent'?

stormi:
What's the rationale behind deprecating BIOS boot?

tui/installer/screens.py
rosslagerwall: This text more or less came from docs, but that sounds better to me so I'll change it.

rosslagerwall:
> What's the rationale behind deprecating BIOS boot?

To reduce maintenance and testing effort, especially as hardware increasingly starts dropping support for legacy BIOS boot.

GeraldEV (approved):

PR #102, opened 2024-02-05
Preserve pool certificates across RPUs

psafont:
Otherwise members cannot establish connection to the coordinator after an RPU

There's a decision to be made for user-installed certificates, which are used to connect to appliances. But this functionality has long worked like that, so I'd rather not upset the balance

### DIFF
--- a/upgrade.py
+++ b/upgrade.py
@@ -391,7 +391,7 @@ class ThirdGenUpgrader(Upgrader):
 
     def buildRestoreList(self, src_base):
         self.restore_list += ['etc/xensource/ptoken', 'etc/xensource/pool.conf',
-                              'etc/xensource/xapi-ssl.pem']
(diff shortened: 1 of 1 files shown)


rosslagerwall (approved):

freddy77 (approved):

liulinC:
@psafont  what is the bad side of preserving user-installed certificates for appliance?

psafont:
There are two distinct aspects:
1. This PR does not preserver user-installed certificates, it maintains the previous behaviour.
2. I believe that after an RPU, the pool needs to be configured to be able to connect to WLB. As part of this configuration, certificates need to be trusted, this is done with user-installed certificates. By preserving certificates, the trust on these certificates is maintained. I don't know whether the whole WLB setup needs to be reconfigured. If it isn't, it seems beneficial to maintain the certificates installed, otherwise the WLB setup might need to be amended to expect that certificates are already installed

PR #104, opened 2024-02-06
CA-388603: Update misleading message (and code)

freddy77:
`testUpgradeForbidden` is called to check we can upgrade. The support for upgrading from MBR was removed and trigger a message.
However the message refers to an utility partition even if this partition is not present.
Looking at code we have the check `utilparts is not None` however `utilparts` contains the list of partition numbers (eventually empty) which is never None (`[]` is not None) so the condition was always True.
Remove the now useless check and update the message to better reflect the current state of code and intention.

### DIFF
--- a/upgrade.py
+++ b/upgrade.py
@@ -209,9 +209,8 @@ class ThirdGenUpgrader(Upgrader):
         primary_fs.unmount()
 
     def testUpgradeForbidden(self, tool):
-        utilparts = tool.utilityPartitions()
-        if tool.partTableType == constants.PARTITION_DOS and utilparts is not None:
-            raise RuntimeError("Util partition detected on DOS partition type, upgrade forbidden.")
+        if tool.partTableType == constants.PARTITION_DOS:
+            raise RuntimeError("Upgrade from a DOS partition type is not supported.")
 
     prepTargetStateChanges = []
     prepTargetArgs = ['primary-disk', 'target-boot-mode', 'boot-partnum', 'primary-partnum', 'logs-partnum', 'swap-partnum', 'storage-partnum']


upgrade.py
```diff
@@ -209,9 +209,8 @@ def __init__(self, source):
         primary_fs.unmount()
 
     def testUpgradeForbidden(self, tool):
-        utilparts = tool.utilityPartitions()
-        if tool.partTableType == constants.PARTITION_DOS and utilparts is not None:
-            raise RuntimeError("Util partition detected on DOS partition type, upgrade forbidden.")
+        if tool.partTableType == constants.PARTITION_DOS:
```
rosslagerwall, line 212: We don't in general support upgrade from DOS partitions.

As far as I know, we also don't support upgrading GPT partition tables to the new layout if the old one had a utility partition. Was that a possible configuration? If yes, then we should check for it and block it.

andyhhp: The old utility partition was one of the things which blocked us using GPT layout.

freddy77: Yes, but the code changed in the meantime and code to upgrade from MBR was removed. If we consider MBR/GPT and utility present/not present the combinations are

1. MBR and not utility partition. Not supported, code removed
2. MBR and utility partition. Not supported, never supported
3. GPT and not utility partition. Working
4. GPT and utility partition. Not clear if this is supported or not

So I don't see why not being clear with point 4 is a stopover for this merge. I would open an issue to test for GPT + utility  but it seems a further improve more than something related to this change (beside the hunk of code affecting)

rosslagerwall (approved):

upgrade.py
rosslagerwall: There's more subtlety to the combinations since you can have different layouts with GPT.

In any case, upgrade from MBR with/without a utility partition to GPT and the new layout is now supported.
Upgrade from GPT (new layout) with or without a utility partition should work too.

freddy77: BTW checked manually and:

-  installing with an utility partition is supported
- upgrading with an utility partition is supported

PR #105, opened 2024-02-12
CP-45747 / CP-47742 4kn merge

rdn32:
Merge of 4kn feature branch:
- Detect disks with block size greater than 512 bytes
- Name of SR type capable of using such disks is specified in a feature file.
- Require such an SR type be used if available and required given the disks to use for storage.
- Expectations for such an SR type documented in doc/features.txt

### DIFF
--- a/doc/features.txt
+++ b/doc/features.txt
@@ -14,3 +14,16 @@ Currently available feature flags are:
 
     This only impacts the UI, the <source> answerfile construct still
     allows to include supplemental packs without this feature flag.
+
+  large-block-capable-sr-type
+
+    Allow the use of an SR type for local storage that (unlike "lvm" and
+    "ext") is not restricted to using disks with 512 byte blocks. The flag
+    file's content must be the name of the SR type.
+
+    The other expectations for such an SR type are:
+    - Its only non-optional creation parameter (as with "lvm" and "ext") is
+      "device", a comma-separated list devices to use.
+    - It should use an identifying prefix for the names of volume groups that
+      it creates, and this prefix should start "XSLocal". (e.g. if the SR type
+      were called "foo", it might use the prefix "XSLocalFOO".)
--- a/answerfile.py
+++ b/answerfile.py
@@ -343,7 +343,7 @@ class Answerfile:
                                   sr_type_mapping,
                                   default=default_sr_type)
 
-        if sr_type != SR_TYPE_LARGE_BLOCK and len(large_block_disks) > 0:
+        if SR_TYPE_LARGE_BLOCK and sr_type != SR_TYPE_LARGE_BLOCK and len(large_block_disks) > 0:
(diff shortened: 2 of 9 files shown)


doc/features.txt
```diff
@@ -14,3 +14,16 @@ Currently available feature flags are:
 
     This only impacts the UI, the <source> answerfile construct still
     allows to include supplemental packs without this feature flag.
+
+  large-block-capable-sr-type
+
+    Allow the use of an SR type for local storage that (unlike "lvm" and
+    "ext") is not restricted to using disks with 512 blocks. The flag file's
```
MarkSymsCtx, line 21: `512 byte blocks`

MarkSymsCtx (changes requested):
One minor tweak to the documentation.

rosslagerwall:
LGTM aside from Mark's minor docs comment.

rosslagerwall (approved):

doc/features.txt
```diff
@@ -14,3 +14,16 @@ Currently available feature flags are:
...
+
+    Allow the use of an SR type for local storage that (unlike "lvm" and
+    "ext") is not restricted to using disks with 512 byte blocks. The flag
+    file's content must be the name of the SR type.
+
+    The other expectations for such an SR type are:
+    - Its only non-optional creation parameter (as with "lvm" and "ext") is
+      "device", a comma-separated list devices to use.
+    - It should use an identifying prefix for the names of volume groups that
+      it creates, and this prefix should start "XSLocal". (e.g. if the SR type
```
stormi, line 28: If I understand the sentence properly, is *with* missing?
```suggestion
      it creates, and this prefix should start with "XSLocal". (e.g. if the SR type
```

doc/features.txt
```diff
@@ -14,3 +14,16 @@ Currently available feature flags are:
...
+      were called "foo", it might use the prefix "XSLocalFOO".)
```
ydirson, line 29: This hardly qualifies as a "feature flag", limits the capability to support large blocks only to a single SR type other, and forbids EXT and LVM to support it in the future.
What about rather a config file containing list of filesystems?

doc/features.txt
rdn32: I'd say that "start" and "start with" (or even "start off") are equivalent in this context.

MarkSymsCtx (approved):

doc/features.txt
rdn32: If both ext and lvm supported large blocks, then this "flag" could be left unset, and nothing should block their use.

That said, I don't think anything I've done has tied our hands in such a way as would prevent us changing our minds about what to do at a later date.

If the options of what SR types were available and when they could be used were more complicated than lvm & ext + one optional large block SR type then I think there'd need to be a rethink about how the UI for choosing SR type would work.

PR #110, opened 2024-03-05
CP-48077: Support python interpreter in shebang of env

DeliZhangX:
In XS 9 python2 is removed, python is just python3 link. Also remain python interpreter (in shebang) will keep compatibility for scripts, e.g. revert-bootloader.py generated by host-upgrade-plugin for both XS 9 and XS 8.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -27,30 +27,22 @@ def run_script(script, stage, *args):
     util.assertDir(constants.SCRIPTS_DIR)
     fd, local_name = tempfile.mkstemp(prefix=stage, dir=constants.SCRIPTS_DIR)
     try:
+        os.close(fd)
         util.fetchFile(script, local_name)
-
-        # check the interpreter
-        fh = os.fdopen(fd)
-        fh.seek(0)
-        line = fh.readline(40)
-        fh.close()
     except:
         raise RuntimeError("Unable to fetch script %s" % script)
-
-    if not line.startswith('#!'):
-        raise RuntimeError("Missing interpreter in %s." % script)
-    interp = line[2:].split()
-    if interp[0] == '/usr/bin/env':
-        if len (interp) < 2 or interp[1] not in ['python3']:
-            raise RuntimeError("Invalid interpreter %s in %s." % (interp[1], script))
-    elif interp[0] not in ['/bin/sh', '/bin/bash', '/usr/bin/python3']:
-        raise RuntimeError("Invalid interpreter %s in %s." % (interp[0], script))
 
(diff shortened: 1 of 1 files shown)


liulinC:
Note: This is to make `host-upgrade-plugin` compatible with XS8 and XS9.

liulinC (approved):

scripts.py
```diff
@@ -41,7 +41,7 @@ def run_script(script, stage, *args):
         raise RuntimeError("Missing interpreter in %s." % script)
     interp = line[2:].split()
     if interp[0] == '/usr/bin/env':
-        if len (interp) < 2 or interp[1] not in ['python3']:
+        if len (interp) < 2 or interp[1] not in ['python', 'python3']:
```
rosslagerwall, line 44: While this works, I think it would be better to remove this whole section about checking the interpreter. The kernel will fail to execute the file in the unusual case that it can't be executed which makes it just another list of items to keep up to date.

freddy77: I agree. When I saw that code time ago it was my tough too. Also if we want to replace the script with a binary executable I don't see the problem.

freddy77: Maybe however you would need a `try`/`except` around `util.runCmd2` below

DeliZhangX: Thank all! Interpreter check is removed now.
`util.runCmd2` internal already handled exception, so need not to do again.

rosslagerwall: `util.runCmd2` catches exceptions raised by calling `subprocess.Popen` but then it reraises them so I think this code does need to handle the exception.

DeliZhangX: If captured `util.runCmd2` exception, what actions needs to do
- logging and raise again
- or do not raise, but return `rc, out, err`, while what's the preferred values?

Could you show some pseudo-code to let me know your exact point?

rosslagerwall: It shouldn't raise it again since in general (with the exception of answerfile generators), script failures are not fatal and callers do not expect it. Also, `runCmd2()` already logs the error.

I would suggest doing something like this:

```
try:
    rc, out, err = util.runCmd2(...)
except Exception:
    # match shell error code for exec failure
    return 127, b'', b''
```

DeliZhangX: Fixed.

liulinC:
The new update basically follow the suggestions, 
Given we have `universal_newlines=True` in runCmd2, so it make sense to return string instead of bytes.

liulinC (approved):

rosslagerwall (approved):

PR #111, opened 2024-04-01
Fix RuntimeError: dictionary changed size during iteration

acefei:
Fix RuntimeError: dictionary changed size during iteration by creating a copy of the keys using list()

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -697,7 +697,7 @@ def writeGuestDiskPartitions(primary_disk, guest_disks):
             assert gd[:5] == '/dev/'
 
             tool = PartitionTool(gd, constants.PARTITION_GPT)
-            tool.deletePartitions(tool.partitions.keys())
+            tool.deletePartitions(list(tool.partitions.keys()))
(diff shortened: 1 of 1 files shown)


ydirson:
What does that fix? `self.deletePartition()` removes items from `self.partitions`, not from the local `numbers` :thinking:

ydirson:
OK I see now with the unique call site at https://github.com/acefei/host-installer/blob/c4355cdbfcdba9e020cf5f3cb58fa58de5ae50ed/backend.py#L700

The commit message describing the called function would be IMHO more clear if it did not imply information only found of the call site.  Eg. mention that it is allowed to be called with `self.partitions.keys()` which *in python3* is an iterator.

A type annotation would also be useful - that method was obviously written to take a `list` type, and ends up being called with an iterator, so it could also be considered to fix the call site instead?

ydirson (approved):

liulinC (approved):

GeraldEV (approved):

PR #112, opened 2024-04-12
CP-47625: Replace mkinitrd with dracut command

liulinC:
CP-47625: Replace mkinitrd with dracut command
    
    dracut has been upgraded and mkinitrd command has been obsoluted
    Update to use dracut to generate initrd
    
    mkinitrd does not has `--latch` option, even in XS8,
    and `/etc/dracut.conf.d` is always exists as dracut package
    install and own the folder.
    Also clean the dead code in the else branch

See job: 3977688 install succeed

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -853,43 +853,22 @@ def __mkinitrd(mounts, partition, package, kernel_version, fcoe_interfaces):
                              '/etc/init.d/sm-multipath', action]) != 0:
                 raise RuntimeError("Failed to generate multipath configuration")
 
-        # Run mkinitrd inside dom0 chroot
+        # Run dracut inside dom0 chroot
         output_file = os.path.join("/boot", "initrd-%s.img" % kernel_version)
 
         # default to only including host specific kernel modules in initrd
-        if os.path.isdir(os.path.join(mounts['root'], 'etc/dracut.conf.d')):
-            # disable multipath on root partition
-            try:
-                if not isDeviceMapperNode(partition):
-                    f = open(os.path.join(mounts['root'], 'etc/dracut.conf.d/xs_disable_multipath.conf'), 'w')
-                    f.write('omit_dracutmodules+=" multipath "\n')
-                    f.close()
-            except:
-                pass
-        else:
-            args = ['--theme=/usr/share/splash']
+        # disable multipath on root partition
+        try:
+            if not isDeviceMapperNode(partition):
+                f = open(os.path.join(mounts['root'], 'etc/dracut.conf.d/xs_disable_multipath.conf'), 'w')
+                f.write('omit_dracutmodules+=" multipath "\n')
+                f.close()
+        except:
+            pass
 
-            if isDeviceMapperNode(partition):
-                # [multipath-root]: /etc/fstab specifies the rootdev by LABEL so we need this to make sure mkinitrd
-                # picks up the master device and not the slave
-                args.append('--rootdev='+ partition)
-            else:
-                args.append('--without-multipath')
+        cmd = ['dracut', output_file, kernel_version]
 
-            cmd = ['mkinitrd', '--latch']
-            cmd.extend( args )
-            if util.runCmd2(['chroot', mounts['root']] + cmd) != 0:
-                raise RuntimeError("Failed to latch arguments for initrd.")
-
-        cmd = ['new-kernel-pkg', '--install', '--mkinitrd']
-
-        # Save command used to create initrd in <initrd_filename>.cmd
-        cmd_logfile = os.path.join(mounts['root'], output_file[1:] + '.cmd')
-        cmd_fh = open(cmd_logfile, "w")
-        print(' '.join(cmd + ['"$@"', kernel_version]), file=cmd_fh)
-        cmd_fh.close()
-
-        if util.runCmd2(['chroot', mounts['root'], '/bin/sh', output_file + '.cmd']) != 0:
+        if util.runCmd2(['chroot', mounts['root']] + cmd) != 0:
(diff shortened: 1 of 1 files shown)


stephenchengCloud (dismissed):

andyhhp:
One minor point.  The word you want is "obsoleted" not "obsoluted".

backend.py
```diff
@@ -853,43 +853,22 @@ def __mkinitrd(mounts, partition, package, kernel_version, fcoe_interfaces):
...
+                f.close()
+        except:
+            pass
 
-        # Save command used to create initrd in <initrd_filename>.cmd
-        cmd_logfile = os.path.join(mounts['root'], output_file[1:] + '.cmd')
-        cmd_fh = open(cmd_logfile, "w")
-        print(' '.join(cmd + ['"$@"', kernel_version]), file=cmd_fh)
-        cmd_fh.close()
+        cmd = ['dracut', '--no-compress', output_file, kernel_version]
```
freddy77, line 869: Why the usage of `--no-compress` ? Currently we compress it

liulinC: no, we do **NOT** compress it in XS8. (stream).
It looks like compress, but actually not, you can check the result in /boot/initrd*.

Not sure whether it is a mkinitrd bug, 
```
        find . -print0 | cpio --null $cpio_owner_root -H newc -o --quiet | \
            $compress >> "${DRACUT_TMPDIR}/initramfs.img";
```

Does NOT generate gzip file while

```
        find . -print0 | cpio --null $cpio_owner_root -H newc -o --quiet | \
            $compress > "${DRACUT_TMPDIR}/initramfs.img";
``` 
does.

Another reasons  we do not compress it explicitly here, is because 
- The compress does not save much disk,   (around 20M?)
- Some xrtuk-xx-xx machine extremely slow to compress, may caused by not fully enable CPU power,
and finally result in installation timeout. (gzip compress takes about 8 minutes)

backend.py
```diff
@@ -1015,12 +994,21 @@ def prepFallback(mounts, primary_disk, primary_partnum):
     proc_modules.close()
 
     # Generate /boot/initrd-fallback.img.
-    cmd = ['mkinitrd', '--verbose']
-    for mod in modules:
-        cmd.append('--with=%s' % mod)
+    cmd = ['dracut', '--verbose', '--no-compress', '--add-drivers', f'{" ".join(modules)}']
```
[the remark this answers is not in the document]
liulinC: well, my first version is some complex string, and I simplify it to current one.
Yes, it can be just `' '.join()`, good point.

backend.py
DeliZhangX: Checked with Lin, current XS9 initrd image is consistent to XS8 that the first cpio party is not compressed and the second party is compressed.

DeliZhangX (approved):

backend.py
liulinC: @andyhhp  after decompress the result binary, xs9 with `--no-compress` is exactly the same behavior as XS8.

XS8:
```
file initrd-fallback.img 
initrd-fallback.img: ASCII cpio archive (SVR4 with no CRC)

[root@genuk-21-10d ~]# cpio -idv < initrd-fallback.img 
.
early_cpio
kernel
kernel/x86
kernel/x86/microcode
kernel/x86/microcode/AuthenticAMD.bin
kernel/x86/microcode/GenuineIntel.bin
27725 blocks


dd if=initrd-fallback.img of=test.gz bs=512 skip=27725

file test.gz
test.gz: gzip compressed data, from Unix, last modified: Wed Apr 10 06:06:13 2024, max compression
```


XS9:

```
file initrd-fallback.img 
initrd-fallback.img: ASCII cpio archive (SVR4 with no CRC)

 cpio -itv < initrd-fallback.img 
drwxr-xr-x   3 root     root            0 Apr  7 09:19 .
-rw-r--r--   1 root     root            2 Apr  7 09:19 early_cpio
drwxr-xr-x   3 root     root            0 Apr  7 09:19 kernel
drwxr-xr-x   3 root     root            0 Apr  7 09:19 kernel/x86
drwxr-xr-x   2 root     root            0 Apr  7 09:19 kernel/x86/microcode
-rw-r--r--   1 root     root     14117888 Apr  7 09:19 kernel/x86/microcode/GenuineIntel.bin
-rw-r--r--   1 root     root        76166 Apr  7 09:19 kernel/x86/microcode/AuthenticAMD.bin
27725 blocks


 dd if=initrd-fallback.img of=test.gz skip=27725
131398+1 records in
131398+1 records out
67276176 bytes (67 MB, 64 MiB) copied, 0.564197 s, 119 MB/s
[root@genuk-21-09d tmp]# file test.gz
test.gz: gzip compressed data, max compression, from Unix, original size modulo 2^32 214974464
```

andyhhp: This is intentional.  The initrd is generally compressed, but microcode is needed so early on boot that it's prepended uncompressed, so it can be accessed directly out of memory.

As a result the initrd appears to be uncompressed, despite most of it being compressed.

freddy77: Can we use `pigz` to make compression faster if it's an issue?

freddy77:
On the commit message, maybe
```
mkinitrd command has been deprecated in favor of dracut.
Update to use dracut to generate initrd.

mkinitrd does not has `--latch` option, even in XS8, and `/etc/dracut.conf.d`
always exists as dracut package installs and owns the folder.
Also clean the dead code in the else branch.
```

Not an English native speaker either.

backend.py
liulinC: I went through the code of dracut and found it will use `pigz` to accelerate and fallback to `gzip -9`. 
I will give it a try.

liulinC: job 3980464, 3980466 shows it can install succeed. (a little bit slower than no-compress)

freddy77 (approved):

stephenchengCloud (approved):

PR #115, opened 2024-04-22
CA-391027: Unblock disk to allow changing partitions

freddy77:
Using `mount=` option during installation allows to store and use files from a specific device during installation.
It's possible to use a preserved partition in the installation disk, however this currently makes the installer fail to upgrade
as the disk would be blocked so it cannot update partitioning schema in the kernel.
Unmount devices and remount after specific change path to allow to use any partition.

It woks, however

- there's no support for `mount=` syntaxes like `LABEL=` or `VG_xxxx`
- code should be possibly moved from `util.py` to `disktools.py`
- check for disk busy should be more generically done

### DIFF
--- a/util.py
+++ b/util.py
@@ -437,3 +437,68 @@ class URL(object):
             assert self.password is None
 
             return self.url
+
+def readCmdlineMounts():
+    class Mount:
+        def __init__(self, dev, mountpoint, options=None, fstype=None):
+            self.dev = dev
+            self.mountpoint = mountpoint
+            self.options = options
+            self.fstype = fstype
+
+        def __repr__(self):
+            return "Mount(%s, %s, %s, %s)" % (self.dev, self.mountpoint, self.fstype, self.options)
+
+        def mount(self):
+            mount(self.dev, self.mountpoint,
+                  self.options if self.options is None else [self.options], self.fstype)
+
+        def umount(self):
+            umount(self.mountpoint)
+
+    mounts = []
+    with open('/proc/cmdline') as f:
+        for m in [x for x in f.read().rstrip().split(b' ') if x.startswith(b'mount=')]:
+            m = m.split(':')
+            if len(m) < 2:
+                continue
+            dev = m[0][6:]
+            fstype = None
+            options = None
+            if len(m) >= 4:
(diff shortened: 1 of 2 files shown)


stormi:
Would it be possible to describe the context of the change (why it is needed) besides the internal CA reference, in the commit message?

util.py
```diff
@@ -437,3 +437,68 @@ def getPlainURL(self):
...
+def deviceIsBusy(dev):
+    try:
+        os.close(os.open(dev, os.O_EXCL|os.O_RDONLY))
+    except OSError as e:
+        return e.errno == errno.EBUSY
+    except:
+        pass
+    return False
+
+class DeviceUnblock:
```
GeraldEV, line 489: This might make more sense as `DeviceBlocker`, when reading the code in upgrade.py I immediately assumed that DeviceUnblock was performing an operation to unblock the device until I read further

util.py
```diff
@@ -437,3 +437,68 @@ def getPlainURL(self):
             assert self.password is None
 
             return self.url
+
+def readCmdlineMounts():
+    class Mount:
```
GeraldEV, line 442: Would it be better to extract this class such that it's not contained within a method, it might be useful elsewhere?

GeraldEV (changes requested):
I think you're right that this code should live in diskutils.py

freddy77:
> Would it be possible to describe the context of the change (why it is needed) besides the internal CA reference, in the commit message?

Update, thanks

util.py
freddy77: I'm a bit divided on just fix this issue and do it completely and properly. In that case supporting all cases and maybe do everything from python instead of splitting between here and an external script

util.py
freddy77: The intention is more unblocking than blocking

ydirson: `Unmounter` would be more factual?

Also, this really looks like a job for a context manager:
```
with DeviceUnmounter():
    ....
```

But I'm curious about what kind of use cases need to use `mount=` on the system disk.

freddy77:
Closing, not really nice, see new PR

PR #116, opened 2024-04-24
CA-391659: Preserve hidden flag updating GPT partitions

freddy77:
Detect hidden flag and save when reading.
Write the flag when partitions are written.


freddy77:
Superseded

PR #118, opened 2024-04-25
CA-391027: Mount devices specified by mount= only when needed, import other needed changes

freddy77:
Using mount= option during installation allows to store and use files from a specific device during installation.
It's possible to use a preserved partition in the installation disk, however this currently makes the installer fail to upgrade as the disk would be blocked so it cannot update partitioning schema in the kernel.
Mount these devices only on needed paths.
Code for mounting these device is moved from S06mount script (now not used) to installer code.

The series includes some commits from master and a mount improvement commit to simplify executing command in the chroot environment.

Tested:
- local preserved first partition on installation disk;
- added scripts to all phases to check they are executed when devices are mounted;
- `LABEL=` specification, even not correct (mount is ignored);
- upgrade on LV.


startup/preinit
```diff
@@ -50,11 +50,7 @@ for arg in $args; do
       [ -z "$ramdisk_done" ] && /opt/xensource/installer/S05ramdisk start
       ramdisk_done=1
       copy=0;;
-    mount=*)
-      [ -z "$mount_done" ] && /opt/xensource/installer/S06mount start
-      mount_done=1
-      copy=0;;
-    sshpassword=* | start=*)
+    sshpassword=* | start=* | mount=*)
```
rosslagerwall, line 53: It would be better not to list `mount` here, then it will be copied as a command-line argument to the installer (`--mount=...`) which can be handled as part of the argument parsing code rather than frobbing the kernel command-line in readCmdlineMounts().

install.py
```diff
@@ -204,7 +204,10 @@ def go(ui, args, answerfile_address, answerfile_script):
         results['extra-repos'] += extra_repo_defs
         logger.log("Driver repos: %s" % str(results['extra-repos']))
 
+        mounter = disktools.DeviceMounter()
```
rosslagerwall, line 207: Implementing the context manager protocol would make this nicer.

disktools.py
```diff
@@ -1267,3 +1267,60 @@ def getMdNodes():
...
+            self.mountpoint = mountpoint
+            self.options = options
+            self.fstype = fstype
+
+        def __repr__(self):
+            return "Mount(%s, %s, %s, %s)" % (self.dev, self.mountpoint, self.fstype, self.options)
+
+        def mount(self):
+            util.mount(self.dev, self.mountpoint,
+                  self.options if self.options is None else [self.options], self.fstype)
```
rosslagerwall, line 1285: It might be nicer if you update the mount() util function to handle options as either a list of a string.

startup/preinit
```diff
@@ -50,11 +50,7 @@ for arg in $args; do
       [ -z "$ramdisk_done" ] && /opt/xensource/installer/S05ramdisk start
       ramdisk_done=1
       copy=0;;
-    mount=*)
-      [ -z "$mount_done" ] && /opt/xensource/installer/S06mount start
```
rosslagerwall, line 54: Don't you also need to remove the S06mount file?

disktools.py
```diff
@@ -1267,3 +1267,60 @@ def getMdNodes():
...
+                fstype = m[1]
+                options = m[2]
+                mnt = m[3]
+            elif len(m) >= 3:
+                fstype = m[1]
+                mnt = m[2]
+            else:
+                mnt = m[1]
+            if dev.startswith("LABEL="):
+                dev = '/dev/disk/by-label/' + quoteDiskLabel(dev[6:])
```
rosslagerwall, line 1312: Why not call `blkid -t` like the old code did? That way it also handles `UUID=...` as well, like it previously did (and doesn't rely on us getting the arbitrary quoting correct).

backend.py
```diff
@@ -1707,6 +1713,8 @@ def verifyRepos(sources, ui):
         if not repo_good:
             raise RuntimeError("Unable to access repository (%s, %s)" % (i['media'], i['address']))
 
+    mounter.umount()
```
rosslagerwall, line 1716: I think all of these mount/umount calls should be in a `try...finally` (or use a context manager like I suggested elsewhere).

install.py
rosslagerwall, line 207: It seems a bit wasteful to recreate this and reparse arguments every time this is used. Can you use a singleton (or just a module level variable)? If you change the argument handling so that `--mount=` is passed to the installer, it could be created during argument parsing.

startup/preinit
freddy77: It makes sense, it's tedious if you need to test using `shell` option on boot. Surely more coherent. I'll do it.

startup/preinit
freddy77: Well... more than needed it's cleaner, but it's easier from the packaging prospective as it does not need to update the `spec` file. That is I'll do it obviously.

disktools.py
freddy77: I don't think we supposed `UUID=` or if we did the support was removed. The script only check for `LABEL=`. One note is that the label in the kernel will be available after the device can be read; that means that the combination of `LABEL=` and USB still setting is currently (even before this commit) not working.

But probably nice to avoid regression and use `blkid`.

backend.py
freddy77: I'm not really familiar with context manager but I'll look at it (I've done already... only not so often so I don't fully remember, no problems). But not sure how context manager interacts with exceptions... oh... the leave part is called like a `finally` part... got it.

install.py
freddy77: Yes, that for sure. I was going to implement a singleton while adding delay to avoid having the delay every time (yes, I saw you suggested to ignore it as we settle the devices... possibly we need to check if we did the settle before the first `DeviceMounter` usage). I'll add the parse in a single place.

freddy77: done

freddy77: So... done I suppose!

disktools.py
```diff
@@ -1267,3 +1267,76 @@ def getMdNodes():
...
+            self.dev = dev
+            self.mountpoint = mountpoint
+            self.options = options
+            self.fstype = fstype
+
+        def __repr__(self):
+            return "Mount(%s, %s, %s, %s)" % (self.dev, self.mountpoint, self.fstype, self.options)
+
+        def mount(self):
+            util.mount(self.dev, self.mountpoint, self.options, self.fstype)
```
freddy77, line 1284: With the previous shell script errors were just ignored. Should we be consistent and ignore and mount/umount errors?

disktools.py
```diff
@@ -1267,3 +1267,76 @@ def getMdNodes():
...
+        cls.mounts = mounts
+
+    def __init__(self):
+        self.mounts = self.__class__.mounts
+
+    def mount(self):
+        for m in self.mounts:
+            m.mount()
+
+    def __umount(self):
```
rosslagerwall, line 1334: Why not make this public?

backend.py
```diff
@@ -1238,6 +1238,9 @@ def installExtLinux(mounts, disk, location=constants.BOOT_LOCATION_MBR):
 # mounting and unmounting of various volumes
 
 def mountVolumes(primary_disk, boot_partnum, primary_partnum, logs_partnum, cleanup, target_boot_mode):
+    mounter = DeviceMounter()
```
rosslagerwall, line 1241: Does this ever get unmounted?

disktools.py
rosslagerwall: I think ignoring errors would be the wrong thing to do, even if it is a slight change in behaviour.

disktools.py
freddy77: Because is not used outside and force users to use context manager version which is a better way to use it

backend.py
freddy77: No, once you call explicitly `mount` you are done

rosslagerwall: The installer makes some calls to sgdisk after the main RPM installation. Will that not be a problem?

disktools.py
freddy77: Based on comment marking as resolved

backend.py
rosslagerwall: I was looking at `setActiveDiskPartition()`. Even if the kernel does not specifically care about flags changes, gidsk might still call the ioctl to re-read the partition table which would presumably cause a warning?

freddy77: Yes, `sgdisk` print a warning message which is harmless. I tested (much less) the downgrade and it succeeded but the downgrade should use `fdisk` which usually give error for disk busy. I'll test again the downgrade also. Maybe added a log message for `sgdisk` stating that an above warning message is not an issue ?

disktools.py
```diff
@@ -1295,3 +1295,76 @@ def getMdNodes():
...
+    def addMountPoints(cls, arg_mounts):
+        mounts = []
+        for m in arg_mounts:
+            m = m.split(':')
+            if len(m) < 2:
+                continue
+            dev = m[0]
+            fstype = None
+            options = None
+            if len(m) >= 4:
```
GeraldEV, line 1329: Should these cases match the specific number of expected elements and fail if there are more than expected?
`if len(m) == 4: ....` ending with some `else: raise...` to show that input wasn't valid?

restore.py
```diff
@@ -206,6 +206,7 @@ def restore_partitions():
             util.bindMount("/dev", "%s/dev" % dest_fs.mount_point)
             util.bindMount("/sys", "%s/sys" % dest_fs.mount_point)
             util.bindMount("/proc", "%s/proc" % dest_fs.mount_point)
+            util.bindMount("/run", "%s/run" % dest_fs.mount_point)
```
GeraldEV, line 209: Is it not worth using the ChrootMounts context manager for this section too?

disktools.py
freddy77: It was to keep compatibility with the previous code. But I suppose it was done for simplicity only. I'll change it

restore.py
freddy77: Done. Took a bit to test it.

backend.py
freddy77: The downgrade process has no issues

GeraldEV (approved):

rosslagerwall (approved):

PR #121, opened 2024-04-30
Add base test script

freddy77:
This test script comes partially from release/xs8 branch. The intention is to have some common way to launch tests. This allows, for instance, to execute tests from packaging tools (like RPM spec files) using the same code.

### DIFF
--- a/test/test.sh
+++ b/test/test.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+export PYTHONPATH="$(realpath $(dirname $0))":"$(realpath $(dirname $0)/..)":"$PYTHONPATH"
+
+cd "$(realpath $(dirname $0))"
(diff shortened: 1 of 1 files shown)


test/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+export PYTHONPATH="$(realpath $(dirname $0))":"$(realpath $(dirname $0)/..)":"$PYTHONPATH"
+
+cd "$(realpath $(dirname $0))"
+PYTHON=python2
+if  head -1 ../init | grep -q python3; then
+	PYTHON=python3
+fi
+$PYTHON test_errors.py -v
```
GeraldEV, line 10: test_errors.py doesn't exist on master yet, is there a PR/branch to add it and the supporting files?

PR #123, opened 2024-05-01
CA-392421: Support DNF output parsing correctly

freddy77:
DNF output is a bit different from YUM causing the progress bar to not be updated while installing packages.
Update code parsing command output.

### DIFF
--- a/repository.py
+++ b/repository.py
@@ -791,28 +791,33 @@ def installFromYum(targets, mounts, progress_callback, cachedir):
         count = 0
         total = 0
         verify_count = 0
+        progressLine = re.compile('.*?(\d+)/(\d+)$')
         while True:
             line = p.stdout.readline()
             if not line:
                 break
             line = line.rstrip()
             logger.log("DNF: %s" % line)
+            # normalize spaces, they easily change based on indentation
+            line = ' '.join(line.split())
             if line == 'Resolving Dependencies':
                 progress_callback(1)
-            elif line == 'Dependencies Resolved':
+            elif line == 'Dependencies resolved.':
                 progress_callback(3)
-            elif line.startswith('-----------------------------------------'):
-                progress_callback(7)
             elif line == 'Running transaction':
                 progress_callback(10)
-            elif line.endswith(' will be installed') or line.endswith(' will be updated'):
-                total += 1
-            elif line.startswith('  Installing : ') or line.startswith('  Updating : '):
+            elif line.startswith('Installing : ') or line.startswith('Updating : '):
                 count += 1
+                m = progressLine.match(line)
+                if m is not None:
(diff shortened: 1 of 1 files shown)


repository.py
```diff
@@ -791,28 +791,36 @@ def installFromYum(targets, mounts, progress_callback, cachedir):
...
             line = p.stdout.readline()
             if not line:
                 break
             line = line.rstrip()
             logger.log("DNF: %s" % line)
+            norm_line = ' '.join(line.split())
             if line == 'Resolving Dependencies':
                 progress_callback(1)
-            elif line == 'Dependencies Resolved':
+            elif line == 'Dependencies Resolved' or line == 'Dependencies resolved.':
```
rosslagerwall, line 804: Does DNF really use both of these or can we just check for the new one?

repository.py
```diff
@@ -791,28 +791,36 @@ def installFromYum(targets, mounts, progress_callback, cachedir):
...
                 progress_callback(3)
             elif line.startswith('-----------------------------------------'):
```
rosslagerwall, line 806: The code now uses a mixture of `line` and `norm_line`. Can it just use `norm_line` everywhere?

repository.py
```diff
@@ -791,28 +791,36 @@ def installFromYum(targets, mounts, progress_callback, cachedir):
...
                 progress_callback(7)
             elif line == 'Running transaction':
                 progress_callback(10)
             elif line.endswith(' will be installed') or line.endswith(' will be updated'):
```
rosslagerwall, line 810: Does this ever match now or is setting of `total` exclusively handled by the regexp match below?

freddy77: I kept Yum compatibility. Yes, on one side it can be confusing. On the other it makes it easier to move code from one branch to the other. Not sure how much as the `master` code is just using `dnf`.

repository.py
freddy77: The checks with `line` are more strict. Potentially the code could evolve with even a more "normalized" version (for instance making all lower case) so you want to keep 2 versions.

repository.py
freddy77: More compatibility code (support both).

repository.py
GeraldEV: I don't see anywhere that `line` is currently used that doesn't exactly match the value of `norm_line` at the same point, it would be nice to use just one

freddy77:
Remove YUM compatibility, use a single `line` variable

liulinC (approved):

liulinC:
Thanks for fixing this up.

GeraldEV (approved):

rosslagerwall (approved):

PR #125, opened 2024-05-02
setEfiBootEntry: make error messages more useful

ydirson:
There were 2 identical messages, making it hard to guess from a screenshot which exec caused an error.

While at it, the entry-removal code path can be executed more than once, so give a way to identify which entry is involved when the call fails.

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1152,7 +1152,7 @@ def setEfiBootEntry(mounts, disk, boot_partnum, install_type, branding):
 
     # First remove existing entries
     rc, out, err = util.runCmd2(["chroot", mounts['root'], "/usr/sbin/efibootmgr"], True, True)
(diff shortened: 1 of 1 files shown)


freddy77 (approved):

andyhhp (approved):

ydirson:
Would this be considered for backporting in #124, or should I open a separate PR?

freddy77:
> Would this be considered for backporting in #124, or should I open a separate PR?

Even technically including this merge in #124 is not possible, as it's a merge in the other direction (`release/xs8` -> `master`) while you require `master` -> `release/xs8`.

ydirson:
> it's a merge in the other direction

I was likely confused by "backport" to think it was in the other direction ;)

So #127 it is

PR #126, opened 2024-05-03
CA-389863: Speed up installation with newer systemd

rosslagerwall:
Each time, RPM execs a process during the install, it closes all open file descriptors up to the hard limit. The default hard limit for systemd services with newer systemd increased from 4096 to
524288. This caused a regression in installation time as dnf spends almost all the time setting the CLOEXEC bit on non-existent file descriptors.

Fix this by ensuring that /dev, /proc, and /sys are bind mounted into the root before starting installation. That way, RPM can close only the open file descriptors rather than all of the possible fds. This decreases the dnf phase of installation on my test host from 900 seconds to 100 seconds.

As a side effect, it silences various complaints about these filesystems not being mounted, simplifies the code, and avoids pointless mount/unmount cycles.

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -839,43 +839,30 @@ def createDom0DiskFilesystems(install_type, disk, target_boot_mode, boot_partnum
                 mount.unmount()
 
 def __mkinitrd(mounts, partition, package, kernel_version, fcoe_interfaces):
+    if isDeviceMapperNode(partition):
+        # Generate a valid multipath configuration for the initrd
+        action = 'generate-fcoe' if fcoe_interfaces else 'generate-bfs'
+        if util.runCmd2(['chroot', mounts['root'],
+                         '/etc/init.d/sm-multipath', action]) != 0:
+            raise RuntimeError("Failed to generate multipath configuration")
 
+    # Run dracut inside dom0 chroot
+    output_file = os.path.join("/boot", "initrd-%s.img" % kernel_version)
+
+    # default to only including host specific kernel modules in initrd
+    # disable multipath on root partition
     try:
-        util.bindMount('/sys', os.path.join(mounts['root'], 'sys'))
-        util.bindMount('/dev', os.path.join(mounts['root'], 'dev'))
-        util.bindMount('/proc', os.path.join(mounts['root'], 'proc'))
-        util.mount('none', os.path.join(mounts['root'], 'tmp'), None, 'tmpfs')
+        if not isDeviceMapperNode(partition):
+            f = open(os.path.join(mounts['root'], 'etc/dracut.conf.d/xs_disable_multipath.conf'), 'w')
+            f.write('omit_dracutmodules+=" multipath "\n')
+            f.close()
+    except:
+        pass
 
-        if isDeviceMapperNode(partition):
-            # Generate a valid multipath configuration for the initrd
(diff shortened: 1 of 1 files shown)


andyhhp:
:100:   This pointless mount/unmounting has irritated me for 12 years

andyhhp (approved):

freddy77:
Please consider https://github.com/xenserver/host-installer/pull/124.

Why not simply set the limit to the old `1024` value instead? This will speed up all commands doing this crazy not optimized scanning of file descriptors.

Also some commands produce different results, one of them is `dracut` which one path is using.

freddy77:
> 💯 This pointless mount/unmounting has irritated me for 12 years

No, it's not always pointless. Actually I was going to add another for the `dnf` cases, not mounting `/proc` and such causes a lot of warnings and potential issues. Maybe the issue is that once the final partition is formatted and we start installing what's the point of mounting and unmounting every time we execute a command?

liulinC (approved):
It make more sense to add to a sequence before installation as lots of functions needs this. 

If https://github.com/xenserver/host-installer/pull/124 has already been merged to release/xs8,  please also consider it for code consistency.

freddy77:
Why not also using `resource.setrlimit(resource.RLIMIT_NOFILE, (1024, 4096))` at the start of installer to make sure any program won't attempt to use that huge limit?

rosslagerwall:
> Please consider https://github.com/xenserver/host-installer/pull/124.

This should supersede at least part of #124.

> Why not simply set the limit to the old 1024 value instead? This will speed up all commands doing this crazy not optimized scanning of file descriptors.

Anything scanning blindly scanning through file descriptors should be fixed to avoid doing that.

> Also some commands produce different results, one of them is dracut which one path is using. grub is another.

If we didn't previosuly have /proc, /dev, and /sys mounted when commands expected it, then that is a bug and this fixes it.

> Maybe the issue is that once the final partition is formatted and we start installing what's the point of mounting and unmounting every time we execute a command?

Exactly, and this PR fixes that.

andyhhp:
I'm with Ross here.  All commands we run should have proc/dev/sys properly in place, and we should prepare them once at the start of day, not ad-hoc per command.

freddy77:
> I'm with Ross here. All commands we run should have proc/dev/sys properly in place, and we should prepare them once at the start of day, not ad-hoc per command.

I never disagreed on this. I disagree on having 2 different versions, one for XS8 and one for XS9.

rosslagerwall:
> So what about:

That's fine with me.

> In my patch I also added as bind mount the /run directory. I think dracut or dnf was complaining about it. Why don't you add also /run too?

In general, I don't think the /run should be bind mounted into a chroot. It would allow e.g. something inside the croot to connect to the host systemd socket and do questionable things. What was the specific error you saw?

rosslagerwall:
> From installation logs

That specific error should be fixed on master by 0103ecdcd6ae4917d5228f57fd2b44b63cc35f8a.

The code in dracut is bogus anyway... findmnt relies on /proc so if /proc is not mounted it will complain about /run not being mounted. Also, if /proc is mounted but /run is not, it will not fail because `findmnt --target` returns the containing mountpoint and exits with status 0. See https://github.com/dracutdevs/dracut/blob/master/dracut.sh#L1587

PR #128, opened 2024-05-07
CA-392758: Remove Firmware Boot Selected flag check

rosslagerwall:
Previously, the installer asserted that both the Block Valid and Firmware Boot Selected flags were set. On some machines, the Firmware Boot Selected flag is not set.

It is not clear why the installer checks this flag. The spec doesn't define its meaning and other software doesn't seem to check it. The only reference I could find was in iPXE where it is unconditionally set. Therefore, simply ignore the flag.

At the same time, instead of hitting an assertion error if the iBFT block is not marked as valid, just skip it. This allows the installation to continue.

### DIFF
--- a/diskutil.py
+++ b/diskutil.py
@@ -16,6 +16,7 @@ from snackutil import ButtonChoiceWindowEx
 
 use_mpath = False
 CDROM_GET_CAPABILITY = 0x5331
+IBFT_BLOCK_VALID_FLAG = 1 << 0
 
 def mpath_cli_is_working():
     regex = re.compile("switchgroup")
@@ -615,7 +616,9 @@ def setup_ibft_nics():
             nm = f.read().strip()
         with open(os.path.join(e, 'flags'), 'r') as f:
             flags = int(f.read().strip())
-            assert (flags & 3) == 3
+            if (flags & IBFT_BLOCK_VALID_FLAG) == 0:
+                logger.log("Skipping %s not marked as valid" % (e,))
+                continue
 
         if mac not in mac_map:
             raise RuntimeError('Found mac %s in iBFT but cannot find matching NIC' % mac)


freddy77:
I found this for target types:
```
Bit 0 : Block Valid Flag
	0 = no, 1=yes
Bit 1 : Firmware Boot Selected Flag
	0 = no, 1 = yes
Bit 2 : Use Radius CHAP
           0 = no, 1 = yes
Bit 3 : Use Radius rCHAP
           0 = no, 1 = yes

All other bits are reserved and must be 0
```
So the test was checking valid and enabled. I agree on replacing `assert` with `if ... continue`.

rosslagerwall:
> I found this for target types:
> 
> ```
> Bit 0 : Block Valid Flag
> 	0 = no, 1=yes
> Bit 1 : Firmware Boot Selected Flag
> 	0 = no, 1 = yes
> Bit 2 : Use Radius CHAP
>            0 = no, 1 = yes
> Bit 3 : Use Radius rCHAP
>            0 = no, 1 = yes
> 
> All other bits are reserved and must be 0
> ```
> 
> So the test was checking valid and enabled. I agree on replacing `assert` with `if ... continue`.

The code is dealing with the NIC structure, not the target structure. The NIC structure has a slightly different definition:

```
Bit 0 : Block Valid Flag
	0 = no, 1=yes
Bit 1 : Firmware Boot Selected Flag
	0 = no, 1 = yes
Bit 2 : Global / Link Local
            0 = Link Local, 1 = Global
```

My point still stands... In neither case does the standard define the meaning of the flags or how they should be interpreted.

freddy77 (approved):

GeraldEV (approved):

rosslagerwall:
I was finally able to test this by patching iPXE to skip setting the flag in certain circumstances and it worked as expected.

PR #131, opened 2024-05-13
Repartition with one single command

freddy77:
Instead of calling sgdisk for each partition call it with a single command with all partitions.
This makes the operation more atomic and make it faster.


bernhardkaindl (approved):
LGTM

GeraldEV (approved):

PR #132, opened 2024-05-13
Reduce difference between master and release/xs8 branches

freddy77:
Use different ways to reduce differences between the 2 branches
1. Make print statements Python 2/3 compatible
    Add parenthesis where possible, use `import __future__` to get full source compatibility in Python 2.

2. Use universal_newlines option for Popen
     The options is supported by Python 2 too, not only Python 3

3. Force list in some cases
    Python 3 tends to use generators which in some cases changes the results, or simply you want to have a list instead.
    Python 3 code already has this change, in Python 2 is mainly a no-operation so copy to Python 2 to reduce source differences.

4. Use range and items to increase Python 2/3 compatibility
    The difference between xrange vs range and items vs iteritems in Python 2 mainly changes list with iterators.
    This could improve memory usage and performance in case of larger containers but this increase the difference between Python 2 and Python 3 source code.
    We don't have huge containers so use Python 3 syntax to reduce source code difference.

5. Use list comprehension if possible
    Syntax is compatible between Python 2 and Python 3.

6. Repartition with one single command
   This commit was merged recently in master branch

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1431,7 +1431,8 @@ def setRootPassword(mounts, root_pwd):
         cmd = ["/usr/sbin/chroot", mounts["root"], "chpasswd", "-e"]
         pipe = subprocess.Popen(cmd, stdin=subprocess.PIPE,
                                      stdout=subprocess.PIPE,
-                                     close_fds=True)
+                                     close_fds=True,
+                                     universal_newlines=True)
         pipe.communicate('root:%s\n' % root_password)
         assert pipe.wait() == 0
     else:
@@ -1439,7 +1440,8 @@ def setRootPassword(mounts, root_pwd):
         pipe = subprocess.Popen(cmd, stdin=subprocess.PIPE,
                                      stdout=subprocess.PIPE,
                                      stderr=subprocess.PIPE,
-                                     close_fds=True)
+                                     close_fds=True,
+                                     universal_newlines=True)
         pipe.communicate(root_password + "\n")
         assert pipe.wait() == 0
 
--- a/backend.py
+++ b/backend.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: GPL-2.0-only
 
+from __future__ import print_function
+
 import os
 import os.path
 import stat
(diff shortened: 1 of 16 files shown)


bernhardkaindl (approved):
Reviewed all changes visually and verified them using:
- `2to3 -w -f print *.py`
- `2to3 -w -f dict *.py`
- `2to3 -w -f map *.py`
- `2to3 -w -f filter *.py`
- `2to3 -w -f xrange *.py`
- Docs: https://docs.python.org/3/library/2to3.html#fixers

2to3 adds the same changes and many more (those are not needed and hence not done)

Likewise, the universal_newlines commit is correct, it is what expected.
Likewise, the conversions from map() to list comprehensions are good.

Because I saw that cpiofile.py is patched as well, suggestion:

Since 2016, with CP-17850: Remove legacy repository and package code, commit https://github.com/xenserver/host-installer/commit/1c4605abf5389ca9593c05b7f303de94dad44a85 and the move of the host-upgrade plugin into its own repo (it uses xcp.cpiofile now instead) the only user of cpiofile.py is this line:
https://github.com/xenserver/host-installer/blob/master/repository.py#L135
```py
primary_xml = cpiofile._Stream("", "r", "gz", primaryfp, 20*512)
primary_dom = parse(primary_xml)
```
This is not using actual cpio functinality, it just uses it to read/download and gunzip the `primary.xml` data when the minidom.parse() parses it below.

We can delete (permanently terminate) this 2nd copy of `cpiofile.py` (the other is `xcp.cpiofile`) from host-installer by replacing these line with:
```py
primary = BytesIO(primaryfp.read())
primary_xml = gzip.GzipFile("", "r", 9, primary)
primary_dom = parse(primary_xml)
primary.close()
```
That means, with 3 lines of code you can remove cpiofile.py, which is a very complicated piece of code:
```py
wc cpiofile.py
 1839  6326 62295 cpiofile.py
```
When needed, we can still import it from `xcp.cpiofile`: @ydirson and I made it Python2 and Python3-compatible with 74.63% code coverage:
https://coveralls.io/builds/67457698/source?filename=xcp%2Fcpiofile.py

The complete pseudo-testcase (needs installing/removing some dependencies to run) is:
```py
import gzip
from io import BytesIO
from xml.dom.minidom import parse, parseString

import util
from repository import URLAccessor

accessor = URLAccessor(util.URL("https://nginx.org/packages/rhel/5"))
primary_location = "/x86_64/repodata/primary.xml.gz"
primaryfp = accessor.openAddress(primary_location)
primary = BytesIO(primaryfp.read())
primary_xml = gzip.GzipFile("", "r", 9, primary)
primary_dom = parse(primary_xml)
primary.close()
primaryfp.close()
package_names = primary_dom.getElementsByTagName("location")
print(package_names)
```

liulinC (approved):
Static code review looks good, 
Better to do some regression XenRT test (at least install) to verify, as the py2->py3 update are sometimes tricky.

PR #133, opened 2024-05-13
Replace cpiofile._Stream: Use GZipFile(BytesIO(fp.read()) and remove cpiofile.py

bernhardkaindl:
During review of #132, I saw that host-installer still has an old obsolete copy of https://github.com/xenserver/python-libs/blob/master/xcp/cpiofile.py:
- `host-installer/cpiofile.py` has some Python3 updates, but it only uses `cpiofile._Stream()`.
- `xcp/cpiofile.py` is fully Py2/3-compatible: https://coveralls.io/github/xenserver/python-libs

Since 2016, with "CP-17850: Remove legacy repository and package code", commit https://github.com/xenserver/host-installer/commit/1c4605abf5389ca9593c05b7f303de94dad44a85 and the move of the host-upgrade plugin into its own repo (it uses `xcp.cpiofile` now instead) the only user of cpiofile.py is this line:
https://github.com/xenserver/host-installer/blob/master/repository.py#L135
```py
primary_xml = cpiofile._Stream("", "r", "gz", primaryfp, 20*512)
primary_dom = parse(primary_xml)
```
This is not using actual `cpio` functionality, and `primary.xml` is pure XML, there is no actual `cpio` format involved. It just uses `cpiofile._Stream()` to read or download and `gunzip` the `primary.xml` data when the minidom.parse() parses it below.

We can delete (permanently terminate) this 2nd copy of `cpiofile.py` (the other is `xcp.cpiofile`) from host-installer by replacing these line with:
```py
primary = BytesIO(primaryfp.read())
primary_xml = gzip.GzipFile("", "r", 9, primary)
primary_dom = parse(primary_xml)
primary.close()
```
That means, with 3 lines of code you can remove cpiofile.py, which is a very complicated piece of code:
```py
wc cpiofile.py
 1839  6326 62295 cpiofile.py
```
The complete pseudo-testcase (needs installing/removing some dependencies to run) is:
```py
import gzip
from io import BytesIO
from xml.dom.minidom import parse, parseString

import util
from repository import URLAccessor

accessor = URLAccessor(util.URL("https://nginx.org/packages/rhel/5"))
primary_location = "/x86_64/repodata/primary.xml.gz"
primaryfp = accessor.openAddress(primary_location)
primary = BytesIO(primaryfp.read())
primary_xml = gzip.GzipFile("", "r", 9, primary)
primary_dom = parse(primary_xml)
primary.close()
primaryfp.close()
package_names = primary_dom.getElementsByTagName("location")
print(package_names)
```


freddy77:
Going to test it. There's a reference to this file in `LICENSE` file (close to the end), it should be removed too.

repository.py
```diff
@@ -130,14 +130,15 @@ def _parse_repodata(self, accessor):
         repomdfp.close()
 
         primaryfp = accessor.openAddress(primary_location)
-        # Open compressed xml using cpiofile._Stream which is an adapter between CpioFile and a stream-like object.
-        # Useful when specifying the URL for HTTP or FTP repository - A simple GzipFile object will not work in this situation.
-        primary_xml = cpiofile._Stream("", "r", "gz", primaryfp, 20*512)
+        # HTTP/FTP accessors don't implement tell(), so read the data into a buffer
+        primarydata = BytesIO(primaryfp.read())
+        primary_xml = gzip.GzipFile("", "r", 9, primarydata)
```
rosslagerwall, line 135: parse() takes a filename or file-like object. Why not pass the BytesIO directly to parse()?

rosslagerwall (commented):
It looks like the uncompressed primary xml on the XenServer ISO is ~1.5 MiB so decompressing it in its entirety should be fine.

repository.py
freddy77: I suppose for the missing `tell()` method. Or you mean the minidom `parse` function? In this case you need not compressed data

freddy77:
> It looks like the uncompressed primary xml on the XenServer ISO is ~1.5 MiB so decompressing it in its entirety should be fine.

yes, I had the same though, then I considered that DOM have the full parsed file which takes quite more than the simple uncompressed version. Unless we want to handle ZIP bombing attacks (which we don't handle so won't be a regression) it should not be an issue.

freddy77:
Code is working

freddy77 (approved):

repository.py
rosslagerwall: Sorry, I got confused seeing the `9` compression level and thought it was compressing some uncompressed data.

I think it would be less confusing to call it like `gzip.GzipFile("", "r", fileobj=primarydata)` but that is a minor complaint.

rosslagerwall (dismissed):

bernhardkaindl (commented):
@freddy77 updated with these changes (thanks for the review and test):

- Cleaned up to use context managers (where possible).
- added `dom.unlink()` to _"make large amounts of memory available sooner"_:
  - https://docs.python.org/3/library/xml.dom.minidom.html#xml.dom.minidom.Node.unlink:
    - _"Break internal references within the DOM so that it will be garbage collected on versions of Python without cyclic GC. Even when cyclic GC is available, using this can make large amounts of memory available sooner, so calling this on DOM objects as soon as they are no longer needed is good practice. This only needs to be called on the Document object, but may be called on child nodes to discard children of that node."_
  - Note: `dom.unlink()` using a context manager is only implemented in Python3
- Use `GzipFile("", "r", fileobj=fileobj)` (like @rosslagerwall suggested)
- Remove the `cpiofile` MIT licence from the `LICENSE` file (@freddy77)

Like before, only tested using my minimal `test-example` and checked using `mypy`/`pylint`/etc to not have any warnings. It should still work like before, but of course it's better to review and test:

Other than adding the call to `dom.unlink()`, this is effectively the same sequence of events, just formatted more nicely due to the use of the context managers:

```py
        # HTTP/FTP accessors don't implement tell(), so read the data into a buffer
        with BytesIO(primaryfp.read()) as fileobj:
            with gzip.GzipFile("", "r", fileobj=fileobj) as primary_xml:
                dom = parse(primary_xml)
                package_names = dom.getElementsByTagName("location")
                package_sizes = dom.getElementsByTagName("size")
                package_checksums = dom.getElementsByTagName("checksum")
                dom.unlink()
```

repository.py
```diff
@@ -130,14 +130,16 @@ def _parse_repodata(self, accessor):
...
-        primary_xml.close()
+
+        # HTTP/FTP accessors don't implement tell(), so read the data into a buffer
+        with BytesIO(primaryfp.read()) as fileobj:
+            with gzip.GzipFile("", "r", fileobj=fileobj) as primary_xml:
+                dom = parse(primary_xml)
+                package_names = dom.getElementsByTagName("location")
+                package_sizes = dom.getElementsByTagName("size")
+                package_checksums = dom.getElementsByTagName("checksum")
+                dom.unlink()
```
freddy77, line 141: The problem with `dom.unlink()` is that it releases too much memory 😆 
In particular the checksums are empty so installation will fail.

bernhardkaindl: Doh! 😆

I forgot to check that the return values of getElementsByTagName() are still part of the DOM and not copied elements or strings!

`dom.unlink()` removed and updated the commit message accordingly!

repository.py
```diff
@@ -130,14 +130,15 @@ def _parse_repodata(self, accessor):
...
-        primary_xml = cpiofile._Stream("", "r", "gz", primaryfp, 20*512)
-        primary_dom = parse(primary_xml)
-        package_names = primary_dom.getElementsByTagName("location")
-        package_sizes = primary_dom.getElementsByTagName("size")
-        package_checksums = primary_dom.getElementsByTagName("checksum")
-        primary_xml.close()
+
+        # HTTP/FTP accessors don't implement tell(), so read the data into a buffer
+        with BytesIO(primaryfp.read()) as fileobj:
+            with gzip.GzipFile("", "r", fileobj=fileobj) as primary_xml:
```
andyhhp, line 136: All support versions of python support multiple context managers in a singe `with` statement, so you should be able to write:
```python
with BytesIO(primaryfp.read()) as fileobj, \
     gzip.GzipFile("", "r", fileobj=fileobj) as primary_xml:
    dom = parse(...
```

freddy77: I can confirm that this works

freddy77 (dismissed):

repository.py
bernhardkaindl: Thanks, applied with shortening of the internal variables in the context:
- The temporary variables inside the context management can use easy short names
- This makes them more natural to read and understand, as the context is small
- This allows to fit the context managers on a line. It makes it even clearer to read:
```py
# HTTP&FTP accessors don't implement tell(): read xml using BytesIO to gunzip it
with BytesIO(primaryfp.read()) as fp, gzip.GzipFile("", "r", fileobj=fp) as xml:
    dom = parse(xml)
    package_names = dom.getElementsByTagName("location")
    package_sizes = dom.getElementsByTagName("size")
    package_checksums = dom.getElementsByTagName("checksum")
```
Closing comments:
- The shortened variables allow the line to fit into 88 columns
- A width of 88 columns is the default width of black
- With this, this commit is also black-compliant.
  - If we choose to reformat using black or darker (only changed lines) this is ready for it.
Final remark:
- I don't like the `\` notation because:
  - It breaks the flow of reading a statement into two lines that have to be put together.
  - It is not uses by black/darker to break long lines,
    - `black` would end up reformat it in this way to remove the `\`:
```py 
with BytesIO(primaryfp.read()) as fp, gzip.GzipFile(
   "", "r", fileobj=fp
) as primary_xml:
    dom = parse(primary_xml)
    package_names = dom.getElementsByTagName("location")
    package_sizes = dom.getElementsByTagName("size")
    package_checksums = dom.getElementsByTagName("checksum")

```
The short form that I've come up now, is much more readable than both other options to me,
because it almost reads like a sentence (I don't like multiline statements because of that),
and find this to be much more (human)-language-like to read:
```py
with BytesIO(primaryfp.read()) as fp, gzip.GzipFile("", "r", fileobj=fp) as xml:
    dom = parse(xml)
    package_names = dom.getElementsByTagName("location")
    package_sizes = dom.getElementsByTagName("size")
    package_checksums = dom.getElementsByTagName("checksum")

```
Of overall, that's a nice improvement

andyhhp (approved):

freddy77 (approved):

PR #137, opened 2024-05-21
CP-49195: Allows to preserve first partition from MBR layout

freddy77:
This allows to install a clean XenServer 8 using NetScaler platform machines based on MBR partution layout.

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -565,10 +565,9 @@ def partitionTargetDisk(disk, existing, preserve_first_partition, create_sr_part
 
     tool = PartitionTool(disk)
 
-    # Cannot preserve partition for legacy DOS partition table.
+    # Cannot preserve utility partition for legacy DOS partition table.
     if tool.partTableType == constants.PARTITION_DOS :
-        if preserve_first_partition == 'true' or (
-                preserve_first_partition == constants.PRESERVE_IF_UTILITY and tool.utilityPartitions()):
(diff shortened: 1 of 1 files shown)


GeraldEV (approved):

backend.py
```diff
@@ -565,10 +565,9 @@ def partitionTargetDisk(disk, existing, preserve_first_partition, create_sr_part
 
     tool = PartitionTool(disk)
 
-    # Cannot preserve partition for legacy DOS partition table.
+    # Cannot preserve utility partition for legacy DOS partition table.
     if tool.partTableType == constants.PARTITION_DOS :
-        if preserve_first_partition == 'true' or (
-                preserve_first_partition == constants.PRESERVE_IF_UTILITY and tool.utilityPartitions()):
+        if preserve_first_partition == constants.PRESERVE_IF_UTILITY and tool.utilityPartitions():
```
rosslagerwall, line 570: How does this work then? Where does the disk get converted from MBR to GPT?

freddy77: I'll check it. Not sure but I tested and works, either upgrade and doing a clean install.

freddy77: One code is in `backend.py`, specifically `writeDom0DiskPartitions`. `gdisk` will convert automatically to GPT and you don't delete the preserved partition.

The other code is in `upgrade.py`, specifically `convertTarget`.

PR #138, opened 2024-05-22
CP-49641: Ignore errors mounting/unmounting explicit mount points

freddy77:
Code to handle `mount=` options were rewritten adding additional error checking.
However this resulted in a feature regression as this poor error handling is used as a "try next if fails" behavior allowing to specify multiple devices for the same mount point. For instance specifying `mount=/dev/sda2:/mnt mount=/dev/disk/by-label/DEMO:/mnt` for the same `/mnt` mount point the two possible device names were attempted. So instead of giving error log the error and continue to restore the old behavior.

### DIFF
--- a/disktools.py
+++ b/disktools.py
@@ -1357,11 +1357,17 @@ class DeviceMounter:
 
     def mount(self):
         for m in self.mounts:
-            m.mount()
+            try:
+                m.mount()
+            except Exception as e:
+                logger.logException(e)
 
     def __umount(self):
         for m in self.mounts:
-            m.umount()
+            try:
+                m.umount()
+            except Exception as e:
+                logger.logException(e)
 
     def __enter__(self):
         self.mount()


GeraldEV (approved):

rosslagerwall (approved):
I don't really think this is a good idea since it is a caller bug or laziness that is passing bad options to the installer. Why can't the caller be fixed instead?

If this is just to unblock things in the short term, fine...

freddy77:
> I don't really think this is a good idea since it is a caller bug or laziness that is passing bad options to the installer. Why can't the caller be fixed instead?
> 
> If this is just to unblock things in the short term, fine...

I'll open a ticket to remember. My idea for the final version is being able to still specify multiple entries for a single mount point but at least one should succeed. So for instance in something like `mount=/dev/sda1:/mnt/one mount=/dev/sda2:/mnt/one mount=/dev/sda3:/mnt/two` if you have `/dev/sda1` and `/dev/sda3` is fine, if you have `/dev/sda1` and `/dev/sda2` will fail as `/mnt/two` is not satisfied.

PR #139, opened 2024-05-22
Use xl command instead of small inline python

freddy77:
Shorter and not requiring changes between Python 2 and 3.

### DIFF
--- a/startup/preinit
+++ b/startup/preinit
@@ -21,7 +21,7 @@ IFS="
 "
 
 xen_commandline() {
-  python3 -c "import xen.lowlevel.xc as xc; print xc.xc().xeninfo()['xen_commandline']"
+  xl info xen_commandline
 }
 
 # Spaces separate arguments except for spaces enclosed by a pair of single or


freddy77:
Tested. `xl` command is available in both XS7 and XS8 during installation.

bernhardkaindl (approved):
Great idea!

Using `xl info xen_commandline` is much more light-weight than firing up a Python3 instance and the `xc` call (5ms vs. >=21ms)

This is also a fix for a Python3 syntax error in this function:
(Changing this print statement to a function on the move to Python3 was missing)

rosslagerwall:
My understanding is that the `xl` command-line ABI is not a stable interface so relying on it is not ideal. libxc is not stable either, but it would be easier to detect if something has gone wrong. I don't suppose it matters too much either way.

andyhhp:
`xl info xen_commandline` isn't liable to change, even if we don't have a strict API/ABI statement about it.  It is basically `xl info | grep $X` and `xen_commandline` is the way it is because that's how `xend` spelt it, and compatibility is wanted.

What I'm less sure about is how well `xl info` will work in an environment which is lacking both `xenstored` and `init-xen-dom0`.  Traditionally, `xl` used to block unconditionally until it saw signs of life from xenstored and while that will not be coming back, there are no guarantees about other dependencies.

freddy77:
> What I'm less sure about is how well `xl info` will work in an environment which is lacking both `xenstored` and `init-xen-dom0`. Traditionally, `xl` used to block unconditionally until it saw signs of life from xenstored and while that will not be coming back, there are no guarantees about other dependencies.

I suppose if it starts to fail, we'll detect it during installation pretty soon.

PR #140, opened 2024-05-22
Backport fix for CA-392758

rosslagerwall:
Previously, the installer asserted that both the Block Valid and Firmware Boot Selected flags were set. On some machines, the Firmware Boot Selected flag is not set.

It is not clear why the installer checks this flag. The spec doesn't define its meaning and other software doesn't seem to check it. The only reference I could find was in iPXE where it is unconditionally set. Therefore, simply ignore the flag.

At the same time, instead of hitting an assertion error if the iBFT block is not marked as valid, just skip it. This allows the installation to continue.

### DIFF
--- a/diskutil.py
+++ b/diskutil.py
@@ -16,6 +16,7 @@ import time
 from snackutil import ButtonChoiceWindowEx
 
 use_mpath = False
+IBFT_BLOCK_VALID_FLAG = 1 << 0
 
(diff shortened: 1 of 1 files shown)


freddy77 (approved):
It looks familiar!

GeraldEV (approved):

PR #144, opened 2024-06-05
CA-392317: Make sure kernel is up to date using Gdisk

freddy77:
The behavior of gdisk if disk layout cannot be updated is different from fdisk; the former write a warning while the latter returns with error.
This results in possible corruptions during the installation. Check and report the warning as error.

### DIFF
--- a/disktools.py
+++ b/disktools.py
@@ -532,7 +532,8 @@ class PartitionToolBase:
         return int(matches.group(1))
 
     # Private methods:
-    def cmdWrap(self, params):
+    @classmethod
+    def cmdWrap(cls, params):
         rv, out, err = util.runCmd2(params, True, True)
         if rv != 0:
             raise Exception(err)
@@ -895,7 +896,7 @@ class DOSPartitionTool(PartitionToolBase):
 
             # CA-35300: sfdisk doesn't return non-zero when the BLKRRPART ioctl fails
             if 'BLKRRPART: Device or resource busy' in output:
-                raise Exception('The disk appears to be in use and partition changes cannot be applied. Reboot and repeat the installation')
+                raise Exception(constants.PARTITIONING_ERROR)
 
             # Verify the table
             # Ignore warnings about partitions apparently with ends beyond the end of the disk
@@ -1042,7 +1043,7 @@ class GPTPartitionTool(PartitionToolBase):
         except:
             # Ignore error code which results from inconsistent initial state
             pass
-        self.cmdWrap([self.SGDISK, '--mbrtogpt', '--clear', self.device])
+        self.gdiskCheck([self.SGDISK, '--mbrtogpt', '--clear', self.device])
 
         has_esp = False
         for part in table.values():
@@ -1080,7 +1081,7 @@ class GPTPartitionTool(PartitionToolBase):
             if 'partuuid' in part:
                 args += ['--partition-guid=%d:%s' % (num,part['partuuid'])]
         args += [self.device]
-        self.cmdWrap(args)
+        self.gdiskCheck(args)
 
         if isDeviceMapperNode(self.device):
             # Create partitions using device mapper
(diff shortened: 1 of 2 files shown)


disktools.py
```diff
@@ -532,8 +532,16 @@ def partitionNumber(self, partitionDevice):
         return int(matches.group(1))
 
     # Private methods:
-    def cmdWrap(self, params):
+    def cmdWrap(self, params, checkGdisk=False):
         rv, out, err = util.runCmd2(params, True, True)
+        # This check is here to check partitions have been updated in the kernel.
+        # Note that not all operations requires the kernel to be updated, so the flag.
+        # In this respect sgdisk differs from sfdisk (former gives warning latter an error)
```
rosslagerwall, line 539: Is it different? See the code elsewhere in this file for sfdisk:

```
            # CA-35300: sfdisk doesn't return non-zero when the BLKRRPART ioctl fails
            if 'BLKRRPART: Device or resource busy' in output:
                raise Exception('The disk appears to be in use and partition changes cannot be applied. Reboot and repeat the installation')
```

Maybe this should raise a similar exception?

disktools.py
```diff
@@ -532,8 +532,16 @@ def partitionNumber(self, partitionDevice):
         return int(matches.group(1))
 
     # Private methods:
-    def cmdWrap(self, params):
+    def cmdWrap(self, params, checkGdisk=False):
```
rosslagerwall, line 535: Putting sgdisk-specific checks in this library function feels a bit weird given that it is used for calling all sorts of other programs.

disktools.py
freddy77: Probably they fixed `sfdisk`, see
```
# sfdisk -f /dev/sda < out ; echo $?
Checking that no-one is using this disk right now ...
BLKRRPART: Device or resource busy
sfdisk:
This disk is currently in use - repartitioning is probably a bad idea.
Umount all file systems, and swapoff all swap partitions on this disk.
Use the --no-reread flag to suppress this check.


Disk /dev/sda: 13054 cylinders, 255 heads, 63 sectors/track
Old situation:
Units: cylinders of 8225280 bytes, blocks of 1024 bytes, counting from 0

   Device Boot Start     End   #cyls    #blocks   Id  System
/dev/sda1   *      0+   1304    1305-  10482381   1c  Hidden W95 FAT32 (LBA)
/dev/sda2       1305    1827-    523-   4194304   83  Linux
/dev/sda3       1827+   2349-    523-   4194304   83  Linux
/dev/sda4       2349+  13053   10705-  85985234+  8e  Linux LVM
New situation:
Units: sectors of 512 bytes, counting from 0

   Device Boot    Start       End   #sectors  Id  System
/dev/sda1   *        63  20964824   20964762  1c  Hidden W95 FAT32 (LBA)
/dev/sda2      20964825  29353432    8388608  83  Linux
/dev/sda3      29353433  37742040    8388608  83  Linux
/dev/sda4      37742041 209712509  171970469  8e  Linux LVM
Warning: partition 2 does not end at a cylinder boundary
Warning: partition 3 does not start at a cylinder boundary
Warning: partition 3 does not end at a cylinder boundary
Warning: partition 4 does not start at a cylinder boundary
Successfully wrote the new partition table

Re-reading the partition table ...
BLKRRPART: Device or resource busy
sfdisk: The command to re-read the partition table failed.
Run partprobe(8), kpartx(8) or reboot your system now,
before using mkfs

If you created or changed a DOS partition, /dev/foo7, say, then use dd(1)
to zero the first 512 bytes:  dd if=/dev/zero of=/dev/foo7 bs=512 count=1
(See fdisk(8).)
1
```
Yes, same message would be more coherent

disktools.py
freddy77: yes, I'll add a `__gdiskCheck` function instead

rosslagerwall (approved):

GeraldEV (dismissed):

disktools.py
```diff
@@ -1093,6 +1094,20 @@ def utilityPartitions(self):
         return [num for num in self.partitions.keys() if
                 self.partitions[num]['id'] == self.ID_EFI_BOOT and self.partitions[num]['partlabel'] == constants.UTILITY_PARTLABEL]
 
+    @classmethod
+    def gdiskCheck(cls, params):
+        out = cls.cmdWrap(params)
+        # This check is here to check partitions have been updated in the kernel.
+        # Note that not all operations requires the kernel to be updated, so the flag.
```
freddy77, line 1101: I just noted the `, so the flag` part is now obsolete. I'll remove it

GeraldEV (approved):

rosslagerwall (approved):

PR #147, opened 2024-06-25
Handle corrupted GPT data with MBR disks

freddy77:
If a disk have partial GPT data corrupted (this can result from an initial usage of GPT followed by some MBR tool usage not cleaning properly GPT) Linux kernel will handle correctly the disk using MBR data but `(s)gdisk` (and so our installer) will complain.
This can be an issue even with clean installations if we need to preserve a partition.
Use interactive `gdisk` which has an option to consider only MBR data in this case.

The `pexpect` replacement is there to avoid an additional dependency.

### DIFF
--- a/ppexpect.py
+++ b/ppexpect.py
@@ -1,0 +1,87 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+# Poor Python Expect.
+# Simplified expect to run a process
+
+import errno
+import fcntl
+import os
+import select
+import subprocess
+import sys
+import time
+
+__all__ = ['Process', 'TimeoutError']
+
+def set_nonblocking(fd):
+    fl = fcntl.fcntl(fd, fcntl.F_GETFL)
+    fcntl.fcntl(fd, fcntl.F_SETFL, fl | os.O_NONBLOCK)
+
+if sys.version_info[0] < 3:
+    class TimeoutError(OSError):
+        def __init__(self):
+            super(TimeoutError, self).__init__(errno.ETIMEDOUT, None)
+else:
+    TimeoutError = TimeoutError
+
+class Process:
+    def __init__(self, cmd):
+        process = subprocess.Popen(
+            cmd,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            close_fds=True,
+            universal_newlines=True,
+            )
+        set_nonblocking(process.stdout.fileno())
+
+        self.process = process
+
+    def __del__(self):
+        self.close()
+
+    def write(self, msg):
+        stdin = self.process.stdin
+        stdin.write(msg)
+        stdin.flush()
+
+    def expect(self, msg, timeout=1.0):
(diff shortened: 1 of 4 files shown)


ppexpect.py
```diff
@@ -0,0 +1,90 @@
...
+def set_nonblocking(fd):
+    fl = fcntl.fcntl(fd, fcntl.F_GETFL)
+    fcntl.fcntl(fd, fcntl.F_SETFL, fl | os.O_NONBLOCK)
+
+if sys.version_info[0] < 3:
+    class TimeoutError(OSError):
+        def __init__(self):
+            super(TimeoutError, self).__init__(errno.ETIMEDOUT, None)
+else:
+    class TimeoutError(TimeoutError):
```
rosslagerwall, line 25: Why is this redefinition of TimeoutError needed?

ppexpect.py
```diff
@@ -0,0 +1,90 @@
...
+                found = msg in data.decode()
+            else:
+                found = msg.match(data.decode())
+            if found:
+                return found
+
+            data += self.__readTimeout(start, timeout)
+
+    def __readTimeout(self, start, timeout):
+        fd = self.process.stdout.fileno()
```
rosslagerwall, line 64: I don't think you need to do this. Can't you just go straight into the select() and if there is data to read immediately it will fall through to the readFd()?

In particular, if the process is constantly writing, it will never hit the timeout because the call to readFd() is before the timeout check.

ppexpect.py
```diff
@@ -0,0 +1,90 @@
...
+        self.close()
+
+    def write(self, msg):
+        stdin = self.process.stdin
+        stdin.write(msg)
+        stdin.flush()
+
+    def expect(self, msg, timeout=1.0):
+        data = b''
+        start = time.time()
```
rosslagerwall, line 52: I think it would be simpler if you used a deadline instead.

So deadline = time.time() + timeout

Then pass deadline into __readTimeout().

In __readTimeout(), you can check `if now > deadline`, and pass in `deadline - now` into select().

ppexpect.py
freddy77: It's for Python 2/3 compatibility.

ppexpect.py
freddy77: Yes, it makes a difference in case of already expired timeout but for the "poor" implementation I'll remove it, it's fine.

ppexpect.py
freddy77: Yes, similar comment, usually you want to handle timeout <= 0 as no timeout, but for the "poor" implementation not supporting it and using a deadline instead is easier.

ppexpect.py
rosslagerwall: Yeah I can see that. I guess I was wondering why TimeoutError needs to be redefined in the Py3 case when it already exists.

freddy77: Now I got the question. To be exported and used by the test that's not attached at the moment! I'll update to a committable state

rosslagerwall (approved):

PR #148, opened 2024-06-26
CP-27521: Set Partition Labels

GeraldEV:
Set partition labels to make it easier to tell them apart!
This will preserve labels on upgrades, it will only set them when new partitions are created.

Example output after a fresh installation:
```
$ sgdisk -p <disk>
<snip>
Number  Start (sector)    End (sector)  Size       Code  Name
   1        46139392        83888127   18.0 GiB    0700  root
   2         8390656        46139391   18.0 GiB    0700  backup
   3        87033856       268435422   86.5 GiB    8E00  Local SR
   4        83888128        84936703   512.0 MiB   EF02  EPT Boot
   5            2048         8390655   4.0 GiB     0700  logs
   6        84936704        87033855   1024.0 MiB  8200  swap
```

Open to suggestions on alternative names for partitions - these values came from an internal ticket and are up for debate/adjustment.


constants.py
```diff
@@ -98,6 +98,13 @@ def error_string(error, logname, with_hd):
 logsfs_label_prefix = 'logs-'
 logsfs_label = logsfs_label_prefix + '%s'
 
+rootpart_label = "root"
+backuppart_label = "backup"
+lvmpart_label = "Local SR"
```
rosslagerwall, line 103: I think this should be `storagepart_label` since it doesn't necessarily use LVM.

rosslagerwall, line 103: This label might be used on the kernel command-line (and in `/dev/disk/by-partlabel`) so using spaces can make it a bit awkward. I'd suggest `localsr`.

constants.py
```diff
@@ -98,6 +98,13 @@ def error_string(error, logname, with_hd):
...
+bootpart_label = "EPT Boot"
```
rosslagerwall, line 104: I think you mean ESP rather than EPT... and echoing the above comment, I'd suggest simply using `esp`.

disktools.py
```diff
@@ -635,7 +635,8 @@ def createPartition(self, id, sizeBytes=None, number=None, order=None, startByte
             'start': startSector,
             'size': sizeSectors,
             'id': id,
-            'active': active
+            'active': active,
+            'partlabel': label
```
andyhhp, line 639: Leave a trailing comma after label.  It will make the next addition here cleaner.

constants.py
```diff
@@ -98,6 +98,13 @@ def error_string(error, logname, with_hd):
...
+logspart_label = "logs"
+swappart_label = "swap"
```
andyhhp, line 106: These want to be consistently capitalised or not, but as they're names, I'd suggest with a capital letter.

constants.py
andyhhp: except it's an initialism for EFI System Partition so should stay in caps.

constants.py
freddy77: IMHO lower case would be consistent with filesystem labels we already use

rosslagerwall (approved):

freddy77 (changes requested):
Please clean the series

freddy77 (approved):

Issue #150, opened 2024-07-03
[xs8] "restore" with UEFI is broken


ydirson:
In https://github.com/xenserver/host-installer/blob/release/xs8/restore.py#L196-L212 we see `backup_fs` being unmounted and then used, resulting in the crash shown below.

Problem was introduced in v10.10.14 with https://github.com/xenserver/host-installer/commit/e9d686ae17bfcad5383f2d89d6bf848c3abd9a3b#diff-2f9ea690c2cdc310e67485fde6a0d74fd91f25e9ac5f5bc6054961b742de74a8R193 (with a commit that does not exist on `master`)

```
INFO     [2024-07-03 08:01:07] Data restoration complete.  About to re-install bootloader.
INFO     [2024-07-03 08:01:07] Unmounting /tmp/restore-backup-cRo2Wv (force = False)
INFO     [2024-07-03 08:01:07] ran ['/bin/umount', '-d', '/tmp/restore-backup-cRo2Wv']; rc 0
...
INFO     [2024-07-03 08:01:08] INSTALL FAILED.
INFO     [2024-07-03 08:01:08] A fatal exception occurred:
INFO     [2024-07-03 08:01:08] Traceback (most recent call last):
  File "/opt/xensource/installer/install.py", line 269, in go
    restore.restoreFromBackup(backup, progress)
  File "/opt/xensource/installer/restore.py", line 212, in restoreFromBackup
    branding = util.readKeyValueFile(os.path.join(backup_fs.mount_point, constants.INVENTORY_FILE))
  File "/opt/xensource/installer/util.py", line 322, in readKeyValueFile
    f = open(filename, "r")
IOError: [Errno 2] No such file or directory: '/tmp/restore-backup-cRo2Wv/etc/xensource-inventory'
```

freddy77:
Yes, we should probably get the "branding" before unmounting the backup file system.

ydirson:
As noted in #151 this is not the only problem in the UEFI code path.

GeraldEV:
Re-opening as this needs to be resolved in XS8 too

PR #151, opened 2024-07-03
restore: read from backup inventory before unmounting (#150)

ydirson:
This fixes a regression introduced in e9d686ae17bfcad5383f2d89d6bf848c3abd9a3b.

This is not sufficient however to get Restore to work un UEFI, as setEfiBootEntry() assumes `mounts['efi']` is populated, which is not the case (only `backend.mountVolumes()` would set that).

### DIFF
--- a/restore.py
+++ b/restore.py
@@ -192,6 +192,11 @@ def restoreFromBackup(backup, progress=lambda x: ()):
             util.assertDir(os.path.dirname(dst_file))
             boot_config.commit(dst_file)
 
+            # prepare for boot loader restoration, before unmounting backup_fs
+            if boot_config.src_fmt == 'grub2' and efi_boot:
+                branding = util.readKeyValueFile(os.path.join(backup_fs.mount_point, constants.INVENTORY_FILE))
+                branding['product-brand'] = branding['PRODUCT_BRAND']
+
             # repartition if needed
             backup_fs.unmount()
             if restore_partitions:
@@ -209,8 +214,6 @@ def restoreFromBackup(backup, progress=lambda x: ()):
             # restore boot loader
             if boot_config.src_fmt == 'grub2':
                 if efi_boot:
-                    branding = util.readKeyValueFile(os.path.join(backup_fs.mount_point, constants.INVENTORY_FILE))
-                    branding['product-brand'] = branding['PRODUCT_BRAND']
                     backend.setEfiBootEntry(mounts, disk_device, boot_partnum, constants.INSTALL_TYPE_RESTORE, branding)
                 else:
                     if location == constants.BOOT_LOCATION_MBR:


freddy77:
Thanks for working on this.

I was wondering why not doing the same thing we do for `dest_fs`, which is mounting back again.

On the other end the code is repartitioning the disk and mounting destination disk again... but won't the partitions be corrupted at that time?

restore.py
```diff
@@ -192,6 +192,11 @@ def restore_partitions():
             util.assertDir(os.path.dirname(dst_file))
             boot_config.commit(dst_file)
 
+            # prepare for boot loader restoration, before unmounting backup_fs
+            if boot_config.src_fmt == 'grub2' and efi_boot:
+                branding = util.readKeyValueFile(os.path.join(backup_fs.mount_point, constants.INVENTORY_FILE))
+                branding['product-brand'] = branding['PRODUCT_BRAND']
+
             # repartition if needed
             backup_fs.unmount()
```
freddy77, line 201: Maybe just move this line inside the below `if` ?
We are restoring MBR so we won't have EFI to restore.

ydirson: possibly, but that rather looks like some improvement that can wait for later, here I'm solely focused on unbreaking the nominal case

ydirson:
> I was wondering why not doing the same thing we do for `dest_fs`, which is mounting back again.
> 
> On the other end the code is repartitioning the disk and mounting destination disk again... but won't the partitions be corrupted at that time?

That's an interesting question. Just wondering about that seems enough to justify reading them before touching the partition table, that looks like the safest move.

The code states:
> Potentially partitions are not created with the same original alignment, but this is not an issue with modern hardware.

Well, if that ever happens with any partition with data, the restored disk is as good as dead anyway.  I'm quite puzzled of how this is managed (recreating a partition table and getting `*fdisk*` to rewrite it all while crossing fingers).  For such things I would rather entrust `python-parted` to do the work and do controlled changes.

This whole repartitioning stuff I find pretty scary TBH - I don't think there is even a safeguard to prevent from attempting to restore from an upgrade made with a different installer version, in which case this code would have no clue it's dealing with something it potentially does not even interpret properly.

restore.py
```diff
@@ -209,8 +214,7 @@ def restore_partitions():
             # restore boot loader
             if boot_config.src_fmt == 'grub2':
                 if efi_boot:
-                    branding = util.readKeyValueFile(os.path.join(backup_fs.mount_point, constants.INVENTORY_FILE))
-                    branding['product-brand'] = branding['PRODUCT_BRAND']
+                    mounts['esp'] = esp
```
GeraldEV, line 217: As you pointed out on https://github.com/xenserver/host-installer/pull/161/files, these solve the same issue
I think adding `esp` to `mounts` should happen when the mounting occurs in case any future changes need to inspect what is mounted and rely on that `mounts` dictionary
With your change here, it would incorrectly appear as if `esp` is not mounted until this code is run
(Another improvement would be to use `'esp' in mounts` instead of `if efi_mounted` but that's for another PR)

GeraldEV:
I've merged https://github.com/xenserver/host-installer/pull/172 which backports the fix from master, this PR will need to be updated to contain only the inventory changes to be considered for merging

PR #153, opened 2024-07-12
CA-395582: Fix installation with multipath enabled

rosslagerwall:
Commit 6495353cce6a ("CA-392317: Make sure kernel is up to date using Gdisk") added a check to ensure that issues with rereading the partition table are detected. However, rereading the partition table always fails when device-mapper is used (BLKRRPART returns EINVAL) which causes spurious installation failures. The host installer correctly uses kpartx to recreate partitions after repartitioning device-mapper devices so simply ignore the warning from sgdisk when using device-mapper.

### DIFF
--- a/disktools.py
+++ b/disktools.py
@@ -1049,7 +1049,11 @@ class GPTPartitionTool(PartitionToolBase):
         except:
             # Ignore error code which results from inconsistent initial state
             pass
-        self.gdiskCheck([self.SGDISK, '--mbrtogpt', '--clear', self.device])
+        args = [self.SGDISK, '--mbrtogpt', '--clear', self.device]
+        if isDeviceMapperNode(self.device):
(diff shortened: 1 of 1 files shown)


MarkSymsCtx (approved):

freddy77:
Looking at similar code for `sfdisk` the implementation is different. Specifically the no check is done in `DOSPartitionTool::writeThisPartitionTable`. I think it would be better to have it done in a more consistent way, it's not that readable a "`check`" function not doing the check also considering that this function purpose is making the command fail as `sfdisk` does.

GeraldEV (approved):

rosslagerwall:
Updated to move the code into writeThisPartitionTable().

freddy77 (approved):

GeraldEV (approved):

PR #155, opened 2024-07-16
Enhance tracing to hunt for restore bug

ydirson:
The particular issue turned out to be that the nested UEFI host being restored had several disks attached.  Since this is done as NVMe namespaces of a single disk, all sharing the same ID, udev in dom0 does not have enough info and messes up /dev/disk/by-id/ on which restore relies because of `/etc/xensource-inventory`:

```
 lrwxrwxrwx 1 root root 13 Jul 15 14:28 nvme-QEMU_NVMe_Ctrl_nvme0 -> ../../nvme0n3
 lrwxrwxrwx 1 root root 15 Jul 15 14:28 nvme-QEMU_NVMe_Ctrl_nvme0-part1 -> ../../nvme0n2p1
 lrwxrwxrwx 1 root root 15 Jul 15 14:28 nvme-QEMU_NVMe_Ctrl_nvme0-part2 -> ../../nvme0n1p2
```

Only the addition of the `tool.dump()` call allowed me to connect the "Failed to create root filesystem" error with that longstanding (in)famous issue, but a few other additions were useful to get there.

### DIFF
--- a/disktools.py
+++ b/disktools.py
@@ -1123,6 +1123,7 @@ def probePartitioningScheme(device):
     if out == 'dos':
         partitionType = constants.PARTITION_DOS
 
+    logger.debug("probePartitioningScheme(%r) => %r", device, partitionType)
     return partitionType
 
 def PartitionTool(device, partitionType=None):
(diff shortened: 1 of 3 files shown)


ydirson:
fixed typos in commit message

freddy77 (approved):

GeraldEV (approved):

PR #157, opened 2024-07-16
[XS8] Enhance tracing to hunt for restore bug

ydirson:
This is a "backport" of #155 to XS8, with one additional safety guard that does not apply to master any more (or more exactly, this is the original work on xs8 that let to the master PR)


ydirson:
Any news on that one?

freddy77 (approved):

PR #161, opened 2024-07-26
`restore.py`/`restoreFromBackup()`: Ensure esp mount point is known before setting EFI boot entries

GeraldEV:
setEfiBootEntry called here: https://github.com/xenserver/host-installer/blob/master/restore.py#L134
Which requires `mounts['esp']`: https://github.com/xenserver/host-installer/blob/master/backend.py#L1136

Manually tested this fixes the regression

```
INFO     [2024-07-26 13:23:01] Traceback (most recent call last):
  File "/opt/xensource/installer/install.py", line 256, in go
    restore.restoreFromBackup(backup, progress)
  File "/opt/xensource/installer/restore.py", line 143, in restoreFromBackup
    backend.setEfiBootEntry(mounts, disk, boot_partnum, constants.INSTALL_TYPE_RESTORE, branding)
  File "/opt/xensource/installer/backend.py", line 1142, in setEfiBootEntry
    if os.path.exists(os.path.join(mounts['esp'], 'EFI/xenserver/shimx64.efi')):
                                   ~~~~~~^^^^^^^
KeyError: 'esp'
```

### DIFF
--- a/restore.py
+++ b/restore.py
@@ -84,10 +84,11 @@ def restoreFromBackup(backup, progress=lambda x: ()):
         dest_fs = util.TempMount(restore_partition, 'restore-dest-')
         efi_mounted = False
         try:
+            mounts = {'root': dest_fs.mount_point, 'boot': os.path.join(dest_fs.mount_point, 'boot')}
             if efi_boot:
-                esp = os.path.join(dest_fs.mount_point, 'boot', 'efi')
-                os.makedirs(esp)
-                util.mount(boot_device, esp)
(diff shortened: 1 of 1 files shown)


bernhardkaindl (approved):

ydirson:
This seems to address the same problem as #151

bernhardkaindl (approved):
@freddy77: @GeraldEV wrote in #165 that the Ext4 `rootfs` PR depends on this PR.

Can you (or @ashwin9390) review it? Thanks!

PR #162, opened 2024-07-30
doc: be explicit about what bool values are accepted

ydirson:


freddy77:
Maybe would be worth stating the comparison is done case-insensitive.

Not a regression, I found a small exception due to code:
```python
    def parseUIConfirmationPrompt(self):
        results = {}
        nodes = getElementsByTagName(self.top_node, ['ui-confirmation-prompt'])
        if len(nodes) > 0:
            results['ui-confirmation-prompt'] = bool(getText(nodes[0]))
        return results
```

The `bool` conversion will return true for any not empty string, even `"false"` or `"0"`.

GeraldEV:
It would be good to address the issue @freddy77 raised to ensure that the docs are correct when this PR is merged

Issue #11, opened 2022-11-21
Adding a safeguard for UEFI/legacy config switch


ydirson:
Currently, when the system is installed using, say, legacy boot, and a user subsequently changes to UEFI for booting a new install media, the choice to upgrade the installed system is still offered, even though it will not work, and will possibly not be detected until late post-installation phase.
Similarly, and possibly more likely, if a backup exists, which can date from before a voluntary switch from legacy to UEFI, it will be offered with similar consequences.

It should be doable to record the boot-type in `/etc/xensource-inventory` so we can avoid early such situations. Another option (or fallback option) could be to rely on the `/etc/fstab` contents to check for an `/boot/efi` entry.

Does it look like something that could be accepted ?

rosslagerwall:
In an ideal world this would be possible but since it might be tricky and introduce a bunch of test edge cases, I think it is OK to block it for now.

There is existing code in the installer which checks for UEFI boot by checking for an ESP like this `efi_boot = boot_part and boot_part['id'] == GPTPartitionTool.ID_EFI_BOOT` so I would suggest using the same pattern.

PR #163, opened 2024-08-07
[XS8] Adding a safeguard for UEFI/legacy config switch

tescande:
Following issue #11, this series adds checks on the current installation and existing backups to not proposed them for upgrade and/or restore if their boot mode (UEFI/legacy) differs from the installer one. To achieve this, a helper function `is_rootf_uefi()` is added.

### DIFF
--- a/product.py
+++ b/product.py
@@ -28,6 +28,18 @@ THIS_PLATFORM_VERSION = Version.from_string(version.PLATFORM_VERSION)
 XENSERVER_7_0_0 = Version([2, 1, 0]) # Platform version
 XENSERVER_MIN_VERSION = XENSERVER_7_0_0
 
+def is_rootfs_uefi(mount_point):
+    try:
+        with open(os.path.join(mount_point, 'etc', 'fstab'), 'r') as fstab:
+            for line in fstab:
+                m = re.search(r'^\s*[^#]+\s/boot/efi\s', line)
+                if m:
+                    return True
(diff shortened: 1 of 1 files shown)


rosslagerwall:
Why block changing the boot mode while upgrading? In principle, there should be no issue with it since the bootloader is reinstalled..

tescande:
This is what the installer already does in [partitionTargetDisk()](https://github.com/xenserver/host-installer/blob/master/backend.py#L545) but a few steps later. So the goal is to not propose an existing installation for upgrade if boot modes mismatch since this will fail anyway with an error message, which is confusing for users when the upgrade seems to have started.

ydirson:
> Why block changing the boot mode while upgrading? In principle, there should be no issue with it since the bootloader is reinstalled.

Isn't there some partitioning differences with that, with BIOS installs lacking a `/boot/EFI` partition?

ydirson:
On `master` this has been superseded by #182, so I'm retargeting this one for xs8 only.

fallen:
ping @ydirson and @tescande is this still relevant? Do we close it or do we have identified a way to make this move forward?

tescande:
Hi @rosslagerwall,
This PR is opened for a very long time and we replied to your comment.
It's been rebased on top of `release/xs8` since this not relevant for master anymore.
Any chance to reconsider it?
egards

PR #165, opened 2024-08-30
IH-416: EXT4 for Dom0 Root & Logs Partitions

GeraldEV:
Add support for EXT4 root and logs partitions on fresh install and upgrade.
Backup and restore is handled by checking the type of the partition being backed-up/restored-to instead of assuming it's EXT3.
This change is intentionally undocumented in doc/answerfile.txt as we are not expecting to support environments which intentionally divert from the default fs type.

This has been tested manually in all ext3/ext4 combinations (e.g. from ext3->ext3, ext3->ext4, etc.) in fresh installs, upgrades, and restores.
A regression test is being introduced into XenRT, see jobs 4087565 & 4088336 and XenRT PR XXX

This PR depends on commit e72687b44b5968aa275f01a44c112e95a28121fb from https://github.com/xenserver/host-installer/pull/161

For XenServer internal, see PRs:
grub.spec - PR 23
host-upgrade-plugin - PR 42
XenRT - PR 10192

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -143,7 +143,7 @@ def getPrepSequence(ans, interactive):
                         progress_scale=100,
                         pass_progress_callback=True))
     seq += [
-        Task(createDom0DiskFilesystems, A(ans, 'install-type', 'primary-disk', 'target-boot-mode', 'boot-partnum', 'primary-partnum', 'logs-partnum', 'disk-label-suffix'), []),
+        Task(createDom0DiskFilesystems, A(ans, 'install-type', 'primary-disk', 'target-boot-mode', 'boot-partnum', 'primary-partnum', 'logs-partnum', 'disk-label-suffix', 'fs-type'), []),
         Task(mountVolumes, A(ans, 'primary-disk', 'boot-partnum', 'primary-partnum', 'logs-partnum', 'cleanup', 'target-boot-mode'), ['mounts', 'cleanup']),
         ]
     return seq
@@ -177,7 +177,7 @@ def getFinalisationSequence(ans):
         Task(writeKeyboardConfiguration, A(ans, 'mounts', 'keymap'), []),
         Task(configureNetworking, A(ans, 'mounts', 'net-admin-interface', 'net-admin-bridge', 'net-admin-configuration', 'manual-hostname', 'manual-nameservers', 'network-hardware', 'preserve-settings', 'network-backend'), []),
         Task(prepareSwapfile, A(ans, 'mounts', 'primary-disk', 'swap-partnum', 'disk-label-suffix'), []),
-        Task(writeFstab, A(ans, 'mounts', 'target-boot-mode', 'primary-disk', 'logs-partnum', 'swap-partnum', 'disk-label-suffix'), []),
+        Task(writeFstab, A(ans, 'mounts', 'target-boot-mode', 'primary-disk', 'logs-partnum', 'swap-partnum', 'disk-label-suffix', 'fs-type'), []),
         Task(enableAgent, A(ans, 'mounts', 'network-backend', 'services'), []),
         Task(configureCC, A(ans, 'mounts'), []),
         Task(writeInventory, A(ans, 'installation-uuid', 'control-domain-uuid', 'mounts', 'primary-disk',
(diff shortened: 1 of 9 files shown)


bernhardkaindl (dismissed):
Sounds/looks good to me - @freddy77

ashwin9390 (approved):
We could consider using pathlib

restore.py
```diff
@@ -69,11 +69,20 @@ def restoreFromBackup(backup, progress=lambda x: ()):
...
+            util.mkfs(restore_fs_type, restore_partition)
         except Exception as e:
             raise RuntimeError("Failed to create root filesystem: %s" % e)
 
+        # format the logs partition if the fs_type is changing
+        logs_partition = partitionDevice(disk, logs_partnum)
+        if restore_fs_type != diskutil.fs_type_from_device(logs_partition):
+            try:
+                util.mkfs(restore_fs_type, logs_partition)
+            except Exception as e:
```
ashwin9390, line 83: We will get generic exception used in static code check ,could we use except OSError as e instead of generic exception.

ashwin9390 (dismissed):

constants.py
```diff
@@ -88,8 +88,8 @@ def error_string(error, logname, with_hd):
 
 # filesystems and partitions types:
 bootfs_type = 'vfat'
-rootfs_type = 'ext3'
-logsfs_type = 'ext3'
+dom0fs_types = ('ext3', 'ext4')
+default_dom0fs_type = 'ext4'
```
freddy77, line 92: I think for compatibility the default type should be `ext3`.

product.py
```diff
@@ -558,7 +560,7 @@ def findXenSourceBackups():
     for p in partitions:
         b = None
         try:
-            b = util.TempMount(p, 'backup-', ['ro'], 'ext3')
+            b = util.TempMount(p, 'backup-', ['ro'], diskutil.fs_type_from_device(p))
```
freddy77, line 563: Why do we even need to pass file system type here? The kernel should be able to detect the correct type.

backend.py
```diff
@@ -803,6 +803,7 @@ def createDom0DiskFilesystems(install_type, disk, target_boot_mode, boot_partnum
     logs_partition = tool.getPartition(logs_partnum)
     if logs_partition:
         run_mkfs = True
+        change_logs_fs_type = diskutil.fs_type_from_device(partitionDevice(disk, logs_partnum)) != fs_type
```
freddy77, line 806: Why `change_logs_fs_type` and not `logs_fs_type` ?

constants.py
freddy77: Why the rename `root` -> `dom0`? I mean, logs and boot are also `dom0` partitions. `root` was more clear to me, it's the partition to be mounted in root (`/`). If `rootfs_type` and `rootfs_types` were too similar I would use `allowed_rootfs_types`.

backend.py
GeraldEV: This is a boolean value to determine if we need to change the fs_type of the logs partition, hence the name
The end of the line has the comparison against the specified fs_type

product.py
GeraldEV: You're correct that the kernel can figure it out for itself but previously it was explicitly stated so we could pick up any inconsistencies between our expectation of the fs type and the actual fs type
I've kept the structure but since we're asking the partition what the fs type is in this code now it can be removed

constants.py
GeraldEV: It was to avoid confusion :D thanks for the suggestion, will update!
Part of the aim of this PR is to allow us to use EXT4 by default in Dom0 for XS9, so making the default EXT3 would nullify that effort

GeraldEV:
Latest commits address comments from the review, happy to squash when everyone has had a chance to check the diff

bernhardkaindl (approved):
@GeraldEV Thanks!

@freddy77 @ashwin9390: It should be easy for you to review the diff (== commit 4)!

The changes are minimal and only what has been asked in the review.

I should mention that @GeraldEV mentioned in the description that this depends on https://github.com/xenserver/host-installer/pull/161, which also needed one more review (but that's done also now!).

ashwin9390 (approved):

freddy77 (approved):
Changes are good to me, but I would merge the `Address comments from PR review`

GeraldEV:
Squashed the review comments commit, merging! :D

PR #167, opened 2024-09-16
Upgrade: forbid upgrading with a key XAPI will reject

ydirson:
XAPI now rejects the default keysize of 7.x era, which must be regenerated before upgrading to 8.3.  Let the installer refuse to initiate a situation where a Rolling Pool Upgrade would be unable to proceed, with not-yet-updated slaves holding the running VMs getting refused connection to the updated part of the pool.

This now stops the upgrade with the following screen, but after letting the user give consent for launching the upgrade:

cert-too-small

### DIFF
--- a/upgrade.py
+++ b/upgrade.py
@@ -4,6 +4,8 @@ import os
 import re
 import shutil
 
+from OpenSSL import crypto
+
 import diskutil
 import product
 from xcp.version import *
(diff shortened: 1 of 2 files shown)


upgrade.py
```diff
@@ -206,11 +208,21 @@ def __init__(self, source):
...
+        self.key_size = cert.get_pubkey().bits()
+        logger.info("ExistingInstallation %s: certificate key size %s", self.key_size)
+
         primary_fs.unmount()
 
     def testUpgradeForbidden(self, tool):
         if tool.partTableType == constants.PARTITION_DOS:
             raise RuntimeError("Upgrade from a DOS partition type is not supported.")
+        if self.key_size < 2048:
+            raise RuntimeError("Server certificate is too small, must regenerate on 8.2.1 before upgrade.")
```
alexbrett, line 225: I'm nervous about stating a specific version number, perhaps better to have this as e.g. `"Current server certificate is too small (%d bits), please regenerate with at least 2048 bits" % (self.key_size,)`?

ydirson: adjusted in last push

ydirson:
any news on this?

freddy77:
What if the certificate is not using RSA ? Other algorithms require smaller key sizes.

stormi:
> What if the certificate is not using RSA ? Other algorithms require smaller key sizes.

Currently, XAPI doesn't support anything but RSA for this certificate.

upgrade.py
```diff
@@ -4,6 +4,8 @@
 import re
 import shutil
 
+from OpenSSL import crypto
```
freddy77, line 7: Do we need an additional dependency in order to use `OpenSSL` module?

psafont:
Xapi currently enforces not only the key size, but the algorithm as well. So to be extra safe I recommend adding a check that the key must be an RSA one. (and change the message accordingly)

upgrade.py
ydirson: This is right.  Is it too big?  I'm not sure we have many alternatives to choose from.

ydirson:
Added the check for RSA type, and made a few improvements

ydirson:
and added back the "before upgrade" that slept away during the message reformulation

ydirson:
Rebased onto current master.  IIUC the only open question was on the use of openssl, waiting for feedback from @freddy77

PR #175, opened 2024-10-28
Dynamically compute layout string based on partition availability

freddy77:
Allows to exclude some partitions from the layout.

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -181,7 +181,8 @@ def getFinalisationSequence(ans):
         Task(enableAgent, A(ans, 'mounts', 'network-backend', 'services'), []),
         Task(configureCC, A(ans, 'mounts'), []),
         Task(writeInventory, A(ans, 'installation-uuid', 'control-domain-uuid', 'mounts', 'primary-disk',
-                               'backup-partnum', 'storage-partnum', 'guest-disks', 'net-admin-bridge',
+                               'backup-partnum', 'logs-partnum', 'boot-partnum', 'swap-partnum', 'storage-partnum',
+                               'guest-disks', 'net-admin-bridge',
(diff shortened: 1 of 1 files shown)


backend.py
```diff
@@ -1570,7 +1572,15 @@ def writeInventory(installID, controlID, mounts, primary_disk, backup_partnum, s
     inv.write("PLATFORM_NAME='%s'\n" % branding['platform-name'])
     inv.write("PLATFORM_VERSION='%s'\n" % branding['platform-version'])
 
-    layout = 'ROOT,BACKUP,LOG,BOOT,SWAP'
+    layout = 'ROOT'
+    if backup_partnum > 0:
```
GeraldEV, line 1576: Does it matter if the partition numbers are in order here? E.g. if backup_partnum was 3, do we need to ensure it appears at position 3 in the resulting list (comma separated list)?

freddy77: I was thinking the same. I just checked, and the actual partition number order is:

1. ROOT
2. BACKUP
3. STORAGE
4. BOOT
5. LOG
6. SWAP

So I have no idea why the order in the `layout` variable is different, but I would keep for compatibility.

The physical order is:

1. LOG
2. BACKUP
3. ROOT
4. BOOT
5. SWAP
6. STORAGE

So, even this, is different from `layout`.

You get the same order with upgrade and clean install.

GeraldEV (approved):

backend.py
rosslagerwall: From what I can find, the only code that uses `PARTITION_LAYOUT` is the host-installer restore code and it doesn't care about the order.

rosslagerwall (approved):

PR #177, opened 2024-11-01
Simplify partition order computation upgrading partition schema

freddy77:
It was not clear the offset.
This will allow to easily avoid some partitions if needed.
Also, is more similar to `writeDom0DiskPartitions` in `backend.py`.


freddy77:
Note that this apply only to `release/xs8` branch, this code is not available in `master`.

rosslagerwall (approved):

PR #186, opened 2024-12-18
CP-52793: Replace sm-multipath from sm

liulinC:
- Drop usage of sm-multipath, configure multipath directly
- Drop fcoe_interfaces support


liulinC:
XenRT test 4180802 looks good.
Note: Install succeed which related to this change. 
There are some failures not relate to this change, may be caused by XenRT hack. (hacked some condition so we got enough machine to run the test)

MarkSymsCtx (dismissed):

backend.py
```diff
@@ -825,13 +824,23 @@ def createDom0DiskFilesystems(install_type, disk, boot_partnum, primary_partnum,
...
+    rv, wwid, err = util.runCmd2(["chroot", mounts["root"], "/usr/lib/udev/scsi_id",
+                               "-g", primary_disk], with_stdout=True, with_stderr=True)
+    if rv != 0:
+        raise RuntimeError("Failed to whitelist %s with error: %s" % (primary_disk, err) )
+
+    # Remove ending line breaker
+    wwid = wwid.strip()
+
+    util.runCmd2(["chroot", mounts["root"], "/usr/sbin/multipath", "-a", wwid],
+                                with_stdout=True, with_stderr=True)
```
freddy77, line 838: Minor: why are you asking for stdout and stderr if you are using none of them? Did you want to check the result?

liulinC: oh, I used to use them, but then I regarding the usage. 
Good point, I will update.

MarkSymsCtx (approved):

PR #188, opened 2025-01-07
CA-404297: Should not bring up extra interfaces

liulinC:
Increase the timeout to 120 seconds, this align with systemd default value

### DIFF
--- a/init
+++ b/init
@@ -61,11 +61,8 @@ def configureNetworking(ui, device, config):
             netcfg[i] = NetInterface(NetInterface.DHCP, nethw[i].hwaddr,
                                      vlan=config_dict['vlan'])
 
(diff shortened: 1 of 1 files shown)


freddy77:
Pretty massive increase I would say. I suppose we have a reason to increase this timeout, like some conditions where 20 seconds is not enough. It would be worth to add a comment on such condition(s).

freddy77:
OT: Would it be sensible to have a kind of splash screen for such long waits?

freddy77 (dismissed):

liulinC:
xenrt: 4191550~4191553, 4191563~4191567

freddy77 (approved):

stephenchengCloud (approved):

PR #189, opened 2025-01-10
CA-402028: Boot xen.pe rather than xen.gz

rosslagerwall:
xen.pe is the new Secure Boot-enabled and signed Xen binary.


MarkSymsCtx (approved):

freddy77:
Minor: I would add in the commit message that it works also for not-secure boot.

freddy77 (approved):

liulinC (approved):

PR #195, opened 2025-02-11
CP-53300: Add 'Memtest86+ (UEFI)' boot menu

liulinC:
memtest86+ is updated to install the kernel into /boot This commit add a conresponding grub menu to load the kernel

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1044,6 +1044,12 @@ def buildBootLoaderMenu(mounts, xen_version, xen_kernel_version, boot_config, se
                                  root=constants.rootfs_label%disk_label_suffix)
         boot_config.append("safe", e)
 
+    e = bootloader.MenuEntry(hypervisor="", hypervisor_args="", kernel="/boot/memtest86+x64.efi",
+                            kernel_args=' '.join([common_kernel_params, "console=tty0", kernel_console_params]),
+                            initrd="", title="Memtest86+ (UEFI)",
+                            root=constants.rootfs_label%disk_label_suffix)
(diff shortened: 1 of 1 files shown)


liulinC:
image

freddy77:
Shouldn't this be conditional on having the file installed?

MarkSymsCtx (approved):

liulinC:
> check

Yes, it can be.
But we customize our dom0 and does not permit customer change, is it necessary to check ?

freddy77:
> > check
> 
> Yes, it can be. But we customize our dom0 and does not permit customer change, is it necessary to check ?

But this project is open source and used by Vates too. It should be a question of some couple lines of code, the file system should be mounted at that stage

ydirson:
We already ship `memtest86+`, no need for complications here just for us :)

freddy77 (approved):

liulinC:
Thanks @freddy77  for the update, 
Note:  `memtest86+` has been updated in XS9 for UEFI,  XCP-ng may want to catch up with this.
There are lots of customization for XS9 that needs to catch up anyway.

stormi:
> Thanks @freddy77 for the update, Note: `memtest86+` has been updated in XS9 for UEFI, XCP-ng may want to catch up with this. There are lots of customization for XS9 that needs to catch up anyway.

We've updated `memtest86+` for UEFI before the XCP-ng 8.3 release anyway.

Regarding XS9, I agree with you, but I'm not aware of any public release of git repositories or SRPMs for XS9 at the moment.

PR #200, opened 2025-02-25
CA-407115: Remove invalid arguments from memtest86+

stephenchengCloud:
xcp/bootloader.py/readGrub2 parses the kernel args for each entries in grub.cfg: 
```
elif l.startswith("linux"):
    kernel, kernel_args = (l.split(None, 1)
                           [1].split(None, 1))
```
Missing args will result in an error.

Added args to make it the same as in XS9 host:/boot/efi/EFi/xenserver/grub.cfg

+++++++++++++++++++++++++++
Updated:
Removed invalid arguments from memtest86+, and raised another PR to handle the case of an entry with no additional arguments

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1045,7 +1045,7 @@ def buildBootLoaderMenu(mounts, xen_version, xen_kernel_version, boot_config, se
         boot_config.append("safe", e)
 
     e = bootloader.MenuEntry(hypervisor="", hypervisor_args="", kernel="/boot/memtest86+x64.efi",
-                            kernel_args=' '.join(["console=tty0", kernel_console_params]),
+                            kernel_args="",
                             initrd="", title="Memtest86+ (UEFI)",
                             root=constants.rootfs_label%disk_label_suffix)
     boot_config.append("memtest", e)


liulinC (dismissed):

bootloader/grub.cfg
```diff
@@ -41,7 +41,7 @@ menuentry "multipath" {
 }
 
 menuentry "memtest" {
-    linux /boot/memtest86+x64.efi
+    linux /boot/memtest86+x64.efi console=tty0 console=hvc0
```
rosslagerwall, line 44: These are not valid memtest86+ arguments. Wouldn't it be better to fix the parsing code to handle the case of an entry with no additional arguments?

(memtest86+ _does_ accept `console` arguments but the format is slightly different according to the docs I can find.)

stephenchengCloud: I copied the args from  XS9 host:/boot/efi/EFi/xenserver/grub.cfg
So we need also to remove the args from this `gurb.cfg` on host.

stephenchengCloud: Rasied another PR in python-libs repo.
@liulinC @rosslagerwall @GeraldEV Could you please review it?
https://github.com/xenserver/python-libs/pull/153

stephenchengCloud: Use this PR to remove invalid arguments from memtest86+.

rosslagerwall (approved):

liulinC (approved):

PR #206, opened 2025-03-11
CP-53844: Fix writing file downloaded from a URL

freddy77:
With Python 3 file data came in binary form.
Open the file we want to write as binary to avoid having to convert binary to character and write data as read from URL.

### DIFF
--- a/repository.py
+++ b/repository.py
@@ -310,7 +310,7 @@ class MainYumRepository(YumRepositoryWithInfo):
             try:
                 infh = self._accessor.openAddress(keyfile)
                 key_path = os.path.join('/root', os.path.basename(keyfile))
-                outfh = open(key_path, "w")
+                outfh = open(key_path, "wb")
                 outfh.write(infh.read())
                 return """
 gpgcheck=1


GeraldEV (approved):

alexbrett (approved):

ydirson:
Is it really a good idea to manipulate a text file as binary?  Isn't whatever problem triggered this PR just a sign that there is a more profound issue?

freddy77:
> Is it really a good idea to manipulate a text file as binary? Isn't whatever problem triggered this PR just a sign that there is a more profound issue?

There's no manipulation, you read a file and write exactly as it was. And this behaviour is consistent with Python 2.

ydirson:
> There's no manipulation, you read a file and write exactly as it was.

For this call, yes, since no handling is done.  Other calls that do parsing of the data ought to use text mode, but the `openAddress` function only ever returns a stream of `bytes` currently.

> And this behaviour is consistent with Python 2.

That is precisely the reason I would not use (and the reason from my comment, even though possibly not in the right PR, as it probably makes sense not to go through conversion to text in this precise case), Python 3 changed distinguished `bytes` from `str` for good reasons.

ydirson:
My feeling was that this change was hiding something worse, and indeed it seems to.  Now, maybe we want to open an issue to discuss it?

freddy77:
> My feeling was that this change was hiding something worse, and indeed it seems to. Now, maybe we want to open an issue to discuss it?

I have no idea what you are meaning. If you want to open an issue feel free but please explain what you are thinking otherwise I cannot see how we can discuss about it.

PR #207, opened 2025-03-11
CA-408187: Parse correctly fdisk output

freddy77:
Newer fdisk output is different potentially causing installer to fail and exit if the disk is still partitioned with MBR schema. This is unlikely but handle it to be able to reinstall a system with such partitioning.

### DIFF
--- a/disktools.py
+++ b/disktools.py
@@ -804,11 +804,12 @@ class DOSPartitionTool(PartitionToolBase):
         for line in out.split("\n"):
             if line == '' or line[0] == '#':
                 pass # Skip comments and blank lines
-            elif state == 0:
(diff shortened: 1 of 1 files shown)


alexbrett (approved):

ydirson:
Those program's output is clearly not meant for parsing.  What about using a python lib instead, like `python-parted`?

freddy77:
> Those program's output is clearly not meant for parsing. What about using a python lib instead, like `python-parted`?

It would make sense, but I would open an issue for tracking, this change fixes an easy reproduce-able problem with a small local fix, rewriting the code is not a small local fix.

liulinC (approved):

PR #209, opened 2025-03-12
CA-408238: Support basic authentication to HTTP server with different port

freddy77:
The second parameter for `add_password` is a partial URL, not a hostname. Setting just the hostname make the authentication apply only if using default port.
Compute a URL with schema and port information.

### DIFF
--- a/util.py
+++ b/util.py
@@ -261,7 +261,7 @@ def fetchFile(source, dest):
             url = URL(source)
             if url.getScheme() in ['http', 'https'] and url.getUsername(): # Auth HTTP(s)
                 passman = urllib2.HTTPPasswordMgrWithDefaultRealm()
-                passman.add_password(None, url.getHostname(), url.getUsername(), url.getPassword())
+                passman.add_password(None, url.getBaseURL(), url.getUsername(), url.getPassword())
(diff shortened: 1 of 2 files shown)


bernhardkaindl (approved):

util.py
```diff
@@ -438,3 +438,10 @@ def getPlainURL(self):
             assert self.password is None
 
             return self.url
+
+    def getBaseURL(self):
+        """Get the URL with schema and network location without username/password."""
+
+        url = self.getPlainURL()
+        parts = urllib.parse.urlsplit(url)
+        return urllib.parse.urlunsplit(parts._replace(path='', query='', fragment=''))
```
rosslagerwall, line 447: I think it would be better do something like this, rather than relying on a private, semi-undocumented method:

```
return urllib.parse.urlunsplit(urllib.parse.SplitResult(parts.scheme, parts.netloc, '', '', ''))
```

rosslagerwall (dismissed):

util.py
freddy77: done

rosslagerwall (approved):

GeraldEV (approved):

PR #213, opened 2025-03-13
Allows invalid UTF-8 output from dnf execution

freddy77:
Dnf output could be intermixed with script output which potentially cause the installation to fail if the output contains invalid UTF-8 encoded characters.
Ignore these lines.

### DIFF
--- a/repository.py
+++ b/repository.py
@@ -797,7 +797,10 @@ def installFromYum(targets, mounts, progress_callback, cachedir):
         verify_count = 0
         progressLine = re.compile('.*?(\d+)/(\d+)$')
         while True:
-            line = p.stdout.readline()
+            try:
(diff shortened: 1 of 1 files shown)


repository.py
```diff
@@ -797,7 +797,10 @@ def installFromYum(targets, mounts, progress_callback, cachedir):
         verify_count = 0
         progressLine = re.compile('.*?(\d+)/(\d+)$')
         while True:
-            line = p.stdout.readline()
+            try:
+                line = p.stdout.readline()
+            except:
```
liulinC, line 802: catch specific exception so we do not ignore other failures by mistake?

bernhardkaindl (approved):

GeraldEV (approved):

PR #218, opened 2025-03-18
[XS8] __mkinitrd: pass -f to dracut

ydirson:
Backport of #152

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -884,7 +884,7 @@ def __mkinitrd(mounts, partition, package, kernel_version, fcoe_interfaces):
     except:
         pass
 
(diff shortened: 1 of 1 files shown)


freddy77 (approved):

bernhardkaindl (approved):
BTW, the description of the original PR was:
> Before https://github.com/xenserver/host-installer/commit/5abd998733a76f3609ca1394e4d923eda8f1a8a3 this code was calling grubby's new-kernel-pkg, which unconditionally passes -f to mkinitrd. Not using -f will cause install failure when a driver disk's rpm's scripts have generated an initrd already (as is the case with the intel igb driver)

PR #222, opened 2025-03-21
Correct name for renamed "fs_type" variable

freddy77:
The variable was renamed in commit 25f4ee21759e3e
(cfr "IH-416, CP-31054, CP-31055: Optional EXT4 Support"). 
This was discovered thanks for instrumentation of Bernhard.


bernhardkaindl (approved):
Nice catch ;)

GeraldEV (approved):
Thanks!

PR #230, opened 2025-03-27
RFC: Mocking the system to test the TUI in developer's environment

ydirson:
This is a tool that, although very imperfect, has already saved me hours in testing developments: it is much easier to check for mistakes, edge cases, etc that way, than remastering an ISO and booting it.

It works by essentially mocking the annoying stuff, and I've been adding to the collection of mocked stuff as needed.

Some problems:
* interrupting it without breaking the tty basically requires to lead the program to a clean exit
* the level of required mocking likely depends on the task at hand, to the current one likely does not fit everyone
* even configuring things like available disks and NICs requires modifying the script, and the level of useful configurability is not clear, because of the above
* the entrypoint of unmodified code was chosen for convenience
* ...

Comments welcome to make this a useful tool, and possibly pave the way to automated TUI testing by mocking inputs


mockrun-installer
```diff
@@ -0,0 +1,506 @@
...
+def main(args):
+    logger.openLog("installer.log", logger.logging.DEBUG)
+    global MOCKDATA
+    MOCKDATA = os.path.join(os.path.dirname(sys.argv[0]), "mockrun-data")
+    logger.log("Starting user interface")
+    tui.init_ui()
+    if False:
+        answerfile = "file:///data/src/host-installer/doc/answerfile.xml"
+    else:
+        answerfile = None
```
freddy77, line 499: That's weird, from the `if False` to the fixed path file.

PR #236, opened 2025-04-07
XSI-1863: Use a disk size of zero if a disk size cannot be determined

GeraldEV:
`getDiskDeviceSize` may fail to find the correct disk size, which could lead to the Python default of `None` being returned - in turn causing larger issues later in the installer where a numeric type is expected

### DIFF
--- a/diskutil.py
+++ b/diskutil.py
@@ -284,6 +284,8 @@ def getDiskDeviceSize(dev):
     elif os.path.exists("/sys/block/%s/size" % dev):
         return int(__readOneLineFile__("/sys/block/%s/size" % dev))
 
+    return 0
+
 def getDiskBlockSize(dev):
     if not dev.startswith("/dev/"):
         dev = '/dev/' + dev


MarkSymsCtx (approved):

freddy77 (approved):
Patch looks good and sensible.

However, OT here. In which cases was the function not detecting the size? Faulty disks? iSCSI having issues? Maybe we should also ignore the disk and not show to the user? Or report the issue to the user?

rosslagerwall (approved):

rosslagerwall:
> However, OT here. In which cases was the function not detecting the size? Faulty disks? iSCSI having issues? Maybe we should also ignore the disk and not show to the user? Or report the issue to the user?

I think the installer already ignores disks that are too small so they should be ignored after this patch.

MarkSymsCtx:
@freddy77 the customer system had a USB SD card reader attached with no medium present

PR #241, opened 2025-04-27
CA-408063: Static IP setting host can't do RPU

liulinC:
The DNS configuration of static IP was saved to /etc/resovl.conf to apply the DNS settings.

In the time of systemd-networkd/systemd-resolved, above config file are owned systemd-resovled and no longer suggested for update

Instead, DNS should be configured in the `Network` section of the .network file for the interface


liulinC:
XenRT test:  suite run 4291019 for RPU with AD joined with static IP
Some failures with unrelated error.

stephenchengCloud (approved):

chunjiez (approved):

PR #243, opened 2025-04-29
CA-409996: Fix KeyError 'target-platform' issue

DeliZhangX:
Define this key early than within answerfile parsing.


DeliZhangX:
https://github.com/xenserver/host-installer/pull/242 is to fix the same issue.

MarkSymsCtx:
> #242 is to fix the same issue.

Do we still need this one then as Frediano's change is merged?

PR #247, opened 2025-05-09
CA-409830: Confirm when overwriting an SR

rosslagerwall:
Add a confirmation dialog to confirm when the user selects a disk that contain an SR since it may result in data being lost.

It applies to the primary disk selection and the guest disk selection.

At the same time, drop the code related to "target-is-sr" since it did determine whether the target is an SR but didn't do anything with the information.

The dialog looks like this:

sr

### DIFF
--- a/diskutil.py
+++ b/diskutil.py
@@ -486,6 +486,10 @@ def log_available_disks():
         if len(dom0disks) == 0:
             logger.log("Unable to find a suitable disk (with a size greater than %dGB) to install to." % constants.min_primary_disk_size)
 
+def isGFS2Filesystem(device):
+    _, out = util.runCmd2(['blkid', '-s', 'TYPE', '-o', 'value', device], with_stdout=True)
+    return out.strip() == 'gfs2'
+
 class Disk:
     def __init__(self, device):
         self.device = device
@@ -499,6 +503,7 @@ class Disk:
 INSTALL_RETAIL = 1
 STORAGE_LVM = 1
 STORAGE_OTHER = 2
+STORAGE_GFS2 = 3
 
 def probeDisk(device):
     """Examines device and reports the apparent presence of a XenServer installation and/or related usage
@@ -509,14 +514,14 @@ def probeDisk(device):
         boot is a tuple of True or False and the partition device
         root is a tuple of None or INSTALL_RETAIL and the partition device
         state is a tuple of True or False and the partition device
-        storage is a tuple of None, STORAGE_LVM or STORAGE_OTHER and the partition device
+        storage is a tuple of None, STORAGE_LVM, STORAGE_GFS2 or STORAGE_OTHER and the partition device
         logs is a tuple of True or False and the partition device
         swap is a tuple of True or False and the partition device
     """
 
     logger.debug("probeDisk(%r)", device)
     disk = Disk(device)
-    possible_srs = []
+    possible_srs = set()
 
     tool = PartitionTool(device)
     tool.dump()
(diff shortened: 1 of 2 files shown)


tui/installer/screens.py
```diff
@@ -32,6 +32,12 @@
 
 MY_PRODUCT_BRAND = PRODUCT_BRAND or PLATFORM_NAME
 
+sr_overwrite_msg = """The selected disk, {}, contains a storage repository. The storage repository may currently be used by other hosts.
+
+Selecting Yes will permanently erase the storage reposory and any virtual machine disks it contains.
```
freddy77, line 37: Mentioned downstream: typo `erase the storage repository`

tui/installer/screens.py
```diff
@@ -613,13 +613,20 @@ def select_primary_disk(answers):
...
-    if 'installation-to-overwrite' in answers:
-        answers['target-is-sr'] = target_is_sr[answers['primary-disk']]
+    # entry contains the 'de' part of the tuple passed in
+    # determine current usage
+    disk = diskutil.probeDisk(entry)
+    if disk.storage[0]:
+        val = snackutil.ButtonChoiceWindowEx(tui.screen,
+            "Confirm Disk Erasure",
+            sr_overwrite_msg.format(diskutil.getHumanDiskLabel(entry, short=True)),
+            ['Yes', 'No'], default=1)
```
freddy77, line 625: Maybe writing a small utility function to reuse these lines? They are the same on below chunk.

freddy77 (approved):

freddy77:
Why not targeting `master` branch first?

GeraldEV (approved):

rosslagerwall:
The updated PR adds to commits to ensure the installer correctly identifies SRs.
It also addresses the comments and skip asking about the primary disk a second time when it is also used for the SR.

freddy77 (approved):

disktools.py
```diff
@@ -185,6 +185,8 @@ def readAllInfo(self):
             if name.startswith('/dev/dm-'):
                 n = int(name[8:])
                 pv['pv_name'] = getDeviceMapperNode(n)
+            elif name.startswith('/dev/disk/by-id/dm-name-'):
```
MarkSymsCtx, line 188: The other option here, not sure if it is possible is to `realpath` the name of the PV.

rosslagerwall: Possibly. That only works if what it is being compared against is also using the canonical path.

Deli has a PR here (https://github.com/xenserver/host-installer/pull/250) to fix this issue by comparing the two paths with `samepath()` since the LVM change affects more than just multipath. I'll probably take that change and drop this commit.

diskutil.py
```diff
@@ -499,6 +503,7 @@ def __init__(self, device):
 INSTALL_RETAIL = 1
 STORAGE_LVM = 1
 STORAGE_OTHER = 2
+STORAGE_GFS2 = 2
```
freddy77, line 506: This is conflicting with `STORAGE_OTHER`

freddy77 (approved):

diskutil.py
```diff
@@ -537,23 +542,24 @@ def probeDisk(device):
...
 
-    lv_tool = len(possible_srs) and LVMTool()
-    for num in possible_srs:
-        part_device = tool._partitionDevice(num)
+    # The entire device may be used as an SR unpartitioned
+    if len(list(tool.items())) == 0:
+        possible_srs.add(device)
 
+    lv_tool = len(possible_srs) and LVMTool()
+    for part_device in possible_srs:
```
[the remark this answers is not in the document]
rosslagerwall: I've added a commit to address this (by raising an exception).

diskutil.py
```diff
@@ -537,23 +542,24 @@ def probeDisk(device):
                 disk.state = (True, part_device)
                 if num + 2 in tool.partitions:
```
GeraldEV, line 543: Is this branch of the code still relevant? Could these SRs still persist on a modern installation? even one which was upgrade from the George release?

GeraldEV (approved):

freddy77 (approved):

PR #248, opened 2025-05-13
CP-54459: Install to SW RAID

GeraldEV:
Create a SW RAID device using mdadm before installation from specified devices
SW RAID installation only available via answerfile and not included in the answerfile documentation as this feature is not designed for wide support - it was specifically requested by an internal customer

UEFI boot does not understand SW RAID and therefore we need to add an EFI boot entry for each physical disk that makes up the SW RAID device. There is a known bug that the GPT header for the second disk in the SW RAID is not correctly populated and therefore its boot entry is invalid - this will be triaged & addressed in a later PR

GRUB already understands Linux SW RAID devices (via its mdraid1x module) which does not need to be explicitly loaded for GRUB to see the SW RAID device - this was verified via the GRUB command line

udev automatically assembles the SW RAID device when starting


backend.py
```diff
@@ -532,6 +537,23 @@ def configureNTP(mounts, ntp_config_method, ntp_servers):
...
+    if os.path.exists(primary_disk):
+        util.runCmd2(['mdadm', '--stop', primary_disk])
+
+    # Zero any superblocks on the physical disks
+    for disk in physical_disks:
+        util.runCmd2(['mdadm', '--zero-superblock', disk])
+
+    # Create multi-device with physical disks
+    # Apply a timeout as the mdadm command can hang waiting for user input in some cases
+    rc = util.runCmd2(['timeout', '60s', 'mdadm', '--create', primary_disk, '--name=XenServer', '--metadata=1.0', '--level=mirror', '--raid-devices=2', physical_disks[0], physical_disks[1]])
```
MarkSymsCtx, line 553: Should this have an assertion that there are precisely only 2 disks in physical_disks?

backend.py
```diff
@@ -105,10 +105,15 @@ def getPrepSequence(ans, interactive):
     seq = [
         Task(util.getUUID, As(ans), ['installation-uuid']),
         Task(util.getUUID, As(ans), ['control-domain-uuid']),
-        Task(util.randomLabelStr, As(ans), ['disk-label-suffix']),
+        Task(util.randomLabelStr, As(ans), ['disk-label-suffix'])]
+
+    if ans['install-type'] == INSTALL_TYPE_FRESH and ans['swraid']:
+        seq.append(Task(setupSWRAIDDevice, A(ans, 'primary-disk', 'physical-disks'), []))
```
rosslagerwall, line 111: I think this needs to move to after `removeBlockingVGs()` otherwise there may be blocking VGs that prevent creating the RAID device.

backend.py
rosslagerwall, line 553: For the customer, I think their desired behaviour is to already have some data (i.e. a utility partition) on one of the disks then once the RAID is created, it will be synced to both disks. Does this handle that correctly? Which disk gets synced to which disk (i.e. is the order important)?

backend.py
```diff
@@ -1180,10 +1202,11 @@ def check_efibootmgr_err(rc, err, install_type, err_type):
         efi = "EFI/xenserver/grubx64.efi"
     else:
         raise RuntimeError("Failed to find EFI loader")
-    rc, err = util.runCmd2(["chroot", mounts['root'], "/usr/sbin/efibootmgr", "-c",
-                            "-L", branding['product-brand'], "-l", '\\' + efi.replace('/', '\\'),
-                            "-d", disk, "-p", str(boot_partnum)], with_stderr=True)
-    check_efibootmgr_err(rc, err, install_type, "Failed to add new efi boot entry")
+    for disk in physical_disks:
```
rosslagerwall, line 1205: I would expect that before the end of installation, the installer should check and wait until the RAID is consistent and fully synchronized. Maybe that would fix the issue writing boot entries?

answerfile.py
```diff
@@ -309,8 +309,17 @@ def parseDisks(self):
                                                              ('no', 'false'),
                                                              ('if-utility', PRESERVE_IF_UTILITY)],
                                                             default='if-utility')
-        primary_disk = normalize_disk(getText(node))
-        results['primary-disk'] = primary_disk
+        results['swraid'] = getBoolAttribute(node, ['swraid'], default=False)
+
+        if results['swraid']:
+            results['primary-disk'] = "/dev/md0"
```
rosslagerwall, line 315: I don't think hardcoding this here is correct. There may be multiple RAID devices and md0 may refer to the wrong one. In particular, I think this customer _will_ have multiple RAID devices.

I think a better approach would be to generate  a unique name for the RAID device here and then you can use that during creation and for the rest of installation. E.g. if you call it `xs-abcdefg` then you can access it via `/dev/md/xs-abcdefg`. Tis name could also be used when referring to an existing installation for upgrade.

(This unique name kind of like the names given to the root filesystem.)

answerfile.py
```diff
@@ -309,8 +309,17 @@ def parseDisks(self):
                                                              ('no', 'false'),
                                                              ('if-utility', PRESERVE_IF_UTILITY)],
                                                             default='if-utility')
-        primary_disk = normalize_disk(getText(node))
-        results['primary-disk'] = primary_disk
+        results['swraid'] = getBoolAttribute(node, ['swraid'], default=False)
```
rosslagerwall, line 312: It might be worth requiring the RAID level to be set in the answerfile and checked here. Even though we currently only plan to support RAID 1, we might extend support in the future.

rosslagerwall:
Will upgrade and restore be handled in a separate PR?

rosslagerwall:
> GRUB already understands Linux SW RAID devices (via its mdraid1x module) which does not need to be explicitly loaded for GRUB to see the SW RAID device - this was verified via the GRUB command line

How did you verify that GRUB was accessing the data via the RAID device rather than just accessing it through one of the RAID members?

rosslagerwall:
`docs/answerfile.txt` needs to be updated. Even if this is not documented in XenServer public docs for customers, it should be documented in the code repository.

answerfile.py
MarkSymsCtx: Maybe also take the name from the answerfile as this customer may want a name other than XenServer.

answerfile.py
GeraldEV: I think we should only introduce switches (like which raid level to use) when they have some effect - so if we decide to add more raid levels it can be added then
I'll specify in the doc that only RAID 1 is available

GeraldEV:
> > GRUB already understands Linux SW RAID devices (via its mdraid1x module) which does not need to be explicitly loaded for GRUB to see the SW RAID device - this was verified via the GRUB command line
> 
> How did you verify that GRUB was accessing the data via the RAID device rather than just accessing it through one of the RAID members?

I entered the grub command line and checked the output of `ls` which showed md entries pointing to the partition(s) we expected

backend.py
GeraldEV: There is an assertion when parsing the answerfile that swraid option specifies two disks, but I'll duplicate it here to protect against future misuse of this method

ydirson:
Nice to see that XenServer is finally interested in this feature - we have an implementation that has been used by XCP-ng users for several years, so maybe we could synchronize? We never opened a PR because XS was not interested at the time, but now it looks like the time can be right.

You may also be interested in #38 as a useful complement.

MarkSymsCtx:
> Nice to see that XenServer is finally interested in this feature - we have an implementation that has been used by XCP-ng users for several years, so maybe we could synchronize? We never opened a PR because XS was not interested at the time, but now it looks like the time can be right.
> 
> You may also be interested in #38 as a useful complement.

We're not going to support this, aside from for one specific internal customer, it will not be available to the general user population.

answerfile.py
rosslagerwall: I'm not completely against that but I'm not sure what the use case would be for that given we don't allow e.g. changing the names of the various dom0 partitions.

backend.py
rosslagerwall: The answerfile code is checking whether the user did something wrong. An assertion here is checking whether the programmer did something wrong...

backend.py
```diff
@@ -532,6 +537,23 @@ def configureNTP(mounts, ntp_config_method, ntp_servers):
...
+    # Stop the multi-device if it exists
+    if os.path.exists(primary_disk):
+        util.runCmd2(['mdadm', '--stop', primary_disk])
+
+    # Zero any superblocks on the physical disks
+    for disk in physical_disks:
+        util.runCmd2(['mdadm', '--zero-superblock', disk])
+
+    # Create multi-device with physical disks
+    # Apply a timeout as the mdadm command can hang waiting for user input in some cases
```
rosslagerwall, line 552: This seems a little odd to me. If mdadm hangs waiting for user input sometimes, shouldn't we handle that rather than just timing out after 60 seconds and raising an error. Or always call mdadm in such a way that it doesn't require user input.

ydirson:
> We're not going to support this, aside from for one specific internal customer, it will not be available to the general user population.

But then, especially if that's clearly not something on your roadmap, why spend time reimplementing something that's been in production on our side for more than 6 years?  We'd be perfectly OK to, for example, get the non-interactive part in shape to be shared, and get the UI part behind a feature flag.

backend.py
GeraldEV: I've updated the SWRAID formation to include a single disk (the first disk in the answerfile), and then add the second disk afterwards - this ensures that the first disk is the "source" and the second disk mirrors it (at least during initial syncing)
Previously it was somewhat random which disk was synced to the other disk

answerfile.py
GeraldEV: Updated to use `/dev/md/xs-<suffix>` during install (although this is just a softlink to the real md device created which has to be used during installation for various annoying reasons)

backend.py
GeraldEV: I've removed the timeout in favour of the cryptic `--run` option which skips the cautions (and possible user input hangs)

backend.py
```diff
@@ -105,24 +105,31 @@ def getPrepSequence(ans, interactive):
...
         ]
 
+    if not interactive:
+        seq.append(Task(verifyRepos, A(ans, 'sources', 'ui'), []))
+
+    if ans['install-type'] == INSTALL_TYPE_FRESH:
+        seq.append(Task(removeBlockingVGs, As(ans, 'guest-disks'), []))
+
+        if ans['swraid']:
+            seq.append(Task(setupSWRAIDDevice, A(ans, 'disk-label-suffix', 'physical-disks'), ['primary-disk']))
```
GeraldEV, line 118: This moved block was so that `setupSWRAIDDevice` could occur before `partitionTargetDisk` (which would otherwise be unhappy with a missing target device) and after `removeBlockingVGs`

backend.py
```diff
@@ -179,7 +187,15 @@ def getRepoSequence(ans, repos):
     return seq
 
 def getFinalisationSequence(ans):
-    seq = [
+    seq = []
+
+    if ans['install-type'] == INSTALL_TYPE_FRESH and ans['swraid']:
+        seq += [
+            Task(addSecondSWRAIDDevice, A(ans, 'primary-disk', 'physical-disks'), []),
```
rosslagerwall, line 194: Wouldn't it be better to add the second device immediately after creating the RAID and then let the sync run at the same time as installation? Otherwise all the data written during installation needs to be resynced at the end of installation when it could just be written to both disks in the first place.

The wait for sync stuff can still live at the end of installation.

backend.py
```diff
@@ -532,6 +548,53 @@ def configureNTP(mounts, ntp_config_method, ntp_servers):
...
+
+    while not isSWRAIDSyncd(primary_disk):
+        time.sleep(constants.swraid_query_interval)
+
+def isSWRAIDSyncd(primary_disk):
+    rc, out = util.runCmd2(['cat', '/sys/block/%s/md/sync_completed' % primary_disk.split("/")[-1]], with_stdout=True)
+    return out.strip() == "none"  # sync_completed shows 'none' when 100% complete
+
+# Wait for SW RAID devices to sync newly written Partition Tables
+def waitForPartitionTableSyncComplete(primary_disk, physical_disks):
```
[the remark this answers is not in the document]
GeraldEV: The device sync only refers to capturing the initial differences between the devices
New writes are not captured in the sync operation, so when the sync has completed there's no guarantee that new writes have completed (the new writes include writing the GPT which must be written before writing the EFI entries)
This generally might be part of a larger issue - discussion on your comment about adding the second device

backend.py
GeraldEV: I've documented some of the investigation on internal ticket CA-410594
In summary, when adding two devices to the SW RAID device before proceeding with the installation; most of the writes to the second device are lost or blocked in some way. It's unclear why this is occurring
The change to add the second device after installation consistently addresses this issue

rosslagerwall: Hmm, I'm not really comfortable with that as a solution without fully understanding the issue. Adding the second disk at the end rather than the beginning is just changing the timing and that is sure to break in unexpected ways at some point in the future depending on the contents and/or speed of the disks.

answerfile.py
```diff
@@ -319,10 +328,10 @@ def parseDisks(self):
         # Guest disk(s) (Local SR)
         guest_disks = set()
         if inc_primary:
-            guest_disks.add(primary_disk)
+            guest_disks.add(results['primary-disk'])
         for node in getElementsByTagName(self.top_node, ['guest-disk']):
             guest_disks.add(normalize_disk(getText(node)))
-        results['sr-on-primary'] = results['primary-disk'] in guest_disks
+        results['sr-on-primary'] = not results['swraid'] and results['primary-disk'] in guest_disks
```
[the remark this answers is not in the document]
GeraldEV: It's not! I made a mistake when checking the requirements! Removed

backend.py
```diff
@@ -1654,13 +1703,59 @@ def touchSshAuthorizedKeys(mounts):
...
+
+    if os.path.exists(os.path.join(mounts['esp'], 'EFI/xenserver/shimx64.efi')):
+        efi = "EFI/xenserver/shimx64.efi"
+    elif os.path.exists(os.path.join(mounts['esp'], 'EFI/xenserver/grubx64.efi')):
+        efi = "EFI/xenserver/grubx64.efi"
+    else:
+        raise RuntimeError("Failed to find EFI loader")
+
+    umountVolumes(mounts, [])
+    util.umount(os.path.join(mounts['root'], 'dev'))
```
GeraldEV, line 1731: These two lines are the ultimate mystery
umountVolumes ends with umount for /tmp/root/dev and /tmp/root, the second of which fails - but then the devtmpfs is re-mounted before it exits.
There's no evidence from lsof of fuser of which processes are using those files/filesystems and adding a generic sleep(x) before the umount doesn't change anything
The only solution I managed to find is that umounting those directories immediately after umountVolumes consistently works and allows the md device to be stopped (and the physical disks syncd)
Ideas (and wild theories) are welcome

backend.py
```diff
@@ -1654,13 +1703,59 @@ def touchSshAuthorizedKeys(mounts):
...
+    try:
+        with open('/sys/block/%s/md/sync_completed' % primary_disk.split("/")[-1]) as sync_file:
+            out = sync_file.read()
+
+        return out.strip() == "none"
+    except Exception as ex:
+        logger.log("Failed to check if SWRAID device is sync'd due to: " + str(ex))
+        return False
+
+def completeSWRAID(mounts, primary_disk, physical_disks, write_boot_entry, install_type, branding, boot_partnum, primary_partnum, logs_partnum, target_boot_mode):
```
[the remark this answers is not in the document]
GeraldEV: The volume (u)mounting is only needed for completing the SWRAID sync, is it better to have a more consistent behaviour in install paths or to avoid unnecessary (u)mounting steps?

We could have the (u)mounting performed within the SWRAID complete section and then install boot entries unconditionally afterwards?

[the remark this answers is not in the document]
rosslagerwall: It should be fine to make the unmount/mount conditional. It should also have a good comment describing why it is needed, since it is not immediately obvious.

backend.py
```diff
@@ -532,6 +550,54 @@ def configureNTP(mounts, ntp_config_method, ntp_servers):
...
+        raise RuntimeError("Failed to add second disk: '%s' to SWRAID device: '%s'" % (physical_disks[1], primary_disk))
+
+    with open('/proc/sys/dev/raid/speed_limit_max', 'w') as speed_file:
+        speed_file.write(str(constants.swraid_speed_write_max))
+
+    primary_disk = os.path.realpath(primary_disk)
+
+    # Update guest-disks
+    if any(disk in physical_disks for disk in guest_disks):
+        guest_disks = list(filter(lambda disk: disk not in physical_disks, guest_disks))
```
rosslagerwall, line 595: I think a list comprehension is preferred in these scenarios. E.g. `guest_disks = [disk for disk in guest_disks if disk not in physical_disks]`

backend.py
```diff
@@ -1185,6 +1240,23 @@ def check_efibootmgr_err(rc, err, install_type, err_type):
...
+def setEfiBootEntry(mounts, disk, boot_partnum, install_type, branding):
+    clearEfiBootEntries(mounts, install_type, branding)
+    addEfiBootEntry(mounts, disk, boot_partnum, install_type, branding)
+
+def writeBootEntries(write_boot_entry, mounts, disks, primary_partnum, boot_partnum, target_boot_mode, install_type, branding, location):
+    for disk in disks:
+        root_partition = partitionDevice(disk, primary_partnum)
+        if target_boot_mode == TARGET_BOOT_MODE_UEFI:
+            if write_boot_entry:
+                clearEfiBootEntries(mounts, install_type, branding)
```
rosslagerwall, line 1252: Shouldn't these calls be replaced by `setEfiBootEntry`?

disktools.py
```diff
@@ -1366,6 +1366,23 @@ def getMdNodes():
...
+    disks = list(map(os.path.basename, physical_disks))
+    try:
+        fh = open('/proc/mdstat')
+        for line in fh:
+            line = line.rstrip()
+            if not ' : ' in line:
+                continue
+            l = line.split(None, 3)
+            if l[2] == 'active' and any(" %s[" % disk in line for disk in disks):
+                nodes.append('/dev/'+l[0])
```
rosslagerwall, line 1380: Spaces around `+` (or use the `getQualifiedDeviceName()` utility function).

disktools.py
```diff
@@ -1366,6 +1366,23 @@ def getMdNodes():
         pass
     return nodes
 
+def getMdDevicesUsing(physical_disks):
+    nodes = []
+    disks = list(map(os.path.basename, physical_disks))
+    try:
+        fh = open('/proc/mdstat')
```
rosslagerwall, line 1373: Use `with open(...` instead to ensure the file is closed if an exception occurs.

disktools.py
```diff
@@ -1366,6 +1366,23 @@ def getMdNodes():
...
+        for line in fh:
+            line = line.rstrip()
+            if not ' : ' in line:
+                continue
+            l = line.split(None, 3)
+            if l[2] == 'active' and any(" %s[" % disk in line for disk in disks):
+                nodes.append('/dev/'+l[0])
+        fh.close()
+    except IOError:
```
rosslagerwall, line 1382: In what cases would this occur? I don't see why this would occur when opening a /proc file.

rosslagerwall (approved):
This basically looks fine to me. A few minor comments.

disktools.py
GeraldEV: This method was mostly derived from `getMdNodes` which is defined above (hence the lack of `with open(` )
I also did not understand the IOError clause but duplicated assuming the writer had a good reason to include it
I'll remove it :D

disktools.py
GeraldEV: Opting for spaces to avoid a circular import with disktools and diskutil

disktools.py
rosslagerwall: Oh, /proc/mdstat will not exist if `md_mod` is not loaded for whatever reason. Maybe that is why the exception is handled.

backend.py
GeraldEV: I think I've introduced a bug during the many iterations
There should be an iteration over the physical disks to add a boot entry for each - I think this used to happen during the "completeSWRAID" phase
It's currently working because it's adding the efi boot entry for the first disk, clearing the entries and then adding it for the second disk (as by this loop)
Where it should instead be clearing it and then adding the entries for each physical disk - I'll do some more dev testing to verify this is the case and create the fix

GeraldEV: I've addressed this in the "Setup the SW RAID device" commit (we clear the efi entries before iterating over the disks to add entries for now)

rosslagerwall (approved):

PR #249, opened 2025-05-14
Upgrade with deprecation of interface-rename

minglumlu:
This PR copes with PR https://github.com/xapi-project/xen-api/pull/6466
The commit 5c4e192 is to resolve the problem when host install is setting up its own networking.
The commit e0ee1e3 is to transform the data of interface-rename to upgraded system.

### DIFF
--- a/netutil.py
+++ b/netutil.py
@@ -371,3 +371,13 @@ def disable_ipv6_module(root):
     dv6fd.write("alias net-pf-10 off\n")
     dv6fd.close()
 
+
+from xcp.net.ifrename.dynamic import DynamicRules
+
+def net_devs_of_last_boot(file_path):
+    dynamic_rules = DynamicRules(file_path)
+    if not dynamic_rules.load_and_parse():
+        LOG.warning(f"Failed to parse the interface-rename dynamic rules.")
+        return None
+    else:
+        return [(d.tname, d.mac.as_string(sep=":")) for d in dynamic_rules.lastboot]
--- a/product.py
+++ b/product.py
@@ -268,9 +268,20 @@ class ExistingInstallation:
                 if not d.get('interfaces') and 'parent' in d:
                     p = fetchIfaceInfoFromNetworkdbAsDict(d['parent'])
                     d['interfaces'] = p['interfaces']
+                    d['hwaddrs'] = p.get('hwaddrs')
 
                 results['net-admin-bridge'] = mgmt_iface
-                results['net-admin-interface'] = d.get('interfaces').split(',')[0]
+                hwaddrs = d.get('hwaddrs')
+                if hwaddrs:
+                    net_admin_ifaces = []
+                    nics = [(name, netutil.getHWAddr(name)) for name in netutil.getNetifList()]
+                    for mac in hwaddrs.lower().split(','):
+                        net_admin_ifaces += [name for (name, hwaddr) in nics if mac == hwaddr]
+                    if len(net_admin_ifaces) == 0:
+                        raise SettingsNotAvailable("Could not find any net admin interface")
+                    results['net-admin-interface'] = net_admin_ifaces[0]
+                else:
+                    results['net-admin-interface'] = d.get('interfaces').split(',')[0]
 
                 if_hwaddr = netutil.getHWAddr(results['net-admin-interface'])
 
--- a/upgrade.py
+++ b/upgrade.py
@@ -503,12 +503,6 @@ class ThirdGenUpgrader(Upgrader):
         print("UPGRADE=true", file=state)
         state.close()
 
-        # The existence of the static-rules.conf is used to detect upgrade from Boston or newer
-        if os.path.exists(os.path.join(mounts['root'], 'etc/sysconfig/network-scripts/interface-rename-data/static-rules.conf')):
-            # CA-82901 - convert any old style ppn referenced to new style ppn references
-            util.runCmd2(['sed', r's/pci\([0-9]\+p[0-9]\+\)/p\1/g', '-i',
-                          os.path.join(mounts['root'], 'etc/sysconfig/network-scripts/interface-rename-data/static-rules.conf')])
-
         net_dict = util.readKeyValueFile(os.path.join(mounts['root'], 'etc/sysconfig/network'))
         if net_dict.get('NETWORKING_IPV6', 'no') == 'no':
             nfd = open(os.path.join(mounts['root'], 'etc/sysconfig/network'), 'a')
--- a/upgrade.py
+++ b/upgrade.py
@@ -487,6 +487,34 @@ class ThirdGenUpgrader(Upgrader):
         return restore_list
 
     completeUpgradeArgs = ['mounts', 'installation-to-overwrite', 'primary-disk', 'backup-partnum', 'logs-partnum', 'net-admin-interface', 'net-admin-bridge', 'net-admin-configuration']
+
+    def convert_interface_rename_data(self, src_file_path, dest_file_path):
+        """
+        Convert data of interface-rename to the data consumed by networkd.
+        """
+        net_devs = netutil.net_devs_of_last_boot(src_file_path)
+        if net_devs is None:
+            logger.error('Failed to parse the interface-rename dynamic rules.')
+            return False
+        else:
+            pattern = r'eth(\d+)'
+            lines = []
+            for (name, mac) in net_devs:
+                match = re.search(pattern, name)
+                if match:
+                    position = match.group(1)
+                    line = f"{position}:mac=\"{mac}\"\n"
+                    logger.log(f"Converted interface-rename data: {line}")
+                    lines += line
+                else:
+                    logger.error(f'Network device name {name} is not like ethN.')
+                    return False
+
+            with open(dest_file_path, mode='w', encoding='utf-8') as f:
+                f.writelines(lines)
+
+            return True
+
     def completeUpgrade(self, mounts, prev_install, target_disk, backup_partnum, logs_partnum, admin_iface, admin_bridge, admin_config):
 
         util.assertDir(os.path.join(mounts['root'], "var/lib/xcp"))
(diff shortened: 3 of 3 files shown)


minglumlu:
Hi @liulinC @DeliZhangX @stephenchengCloud  @robhoes @psafont @lindig @rosslagerwall 
May I please ask your review on this PR?  Thanks.

netutil.py
```diff
@@ -371,3 +371,13 @@ def disable_ipv6_module(root):
     dv6fd.write("alias net-pf-10 off\n")
     dv6fd.close()
 
+
+from xcp.net.ifrename.dynamic import DynamicRules
```
liulinC, line 375: Is it possible/complicated that you do not use DynamicRules here?
The reason why I ask this is because you are now the only user want to use this and we are going to drop this feature from the library.

minglumlu: This is the only source of the last order. So it would have to be parsed either by the existing code, or new code. The existing code can be dropped when the upgrade path is not possible.

product.py
```diff
@@ -268,9 +268,20 @@ def fetchIfaceInfoFromNetworkdbAsDict(bridge, iface=None):
...
 
                 results['net-admin-bridge'] = mgmt_iface
-                results['net-admin-interface'] = d.get('interfaces').split(',')[0]
+                hwaddrs = d.get('hwaddrs')
+                if hwaddrs:
+                    net_admin_ifaces = []
+                    nics = [(name, netutil.getHWAddr(name)) for name in netutil.getNetifList()]
+                    for mac in hwaddrs.lower().split(','):
+                        net_admin_ifaces += [name for (name, hwaddr) in nics if mac == hwaddr]
+                    if len(net_admin_ifaces) == 0:
```
rosslagerwall, line 280: `if not net_admin_ifaces:` is preferred. See PEP8:

```
For sequences, (strings, lists, tuples), use the fact that empty sequences are false:

# Correct:
if not seq:
if seq:

# Wrong:
if len(seq):
if not len(seq):
```

product.py
```diff
@@ -268,9 +268,20 @@ def fetchIfaceInfoFromNetworkdbAsDict(bridge, iface=None):
...
+                    d['hwaddrs'] = p.get('hwaddrs')
 
                 results['net-admin-bridge'] = mgmt_iface
-                results['net-admin-interface'] = d.get('interfaces').split(',')[0]
+                hwaddrs = d.get('hwaddrs')
+                if hwaddrs:
+                    net_admin_ifaces = []
+                    nics = [(name, netutil.getHWAddr(name)) for name in netutil.getNetifList()]
+                    for mac in hwaddrs.lower().split(','):
+                        net_admin_ifaces += [name for (name, hwaddr) in nics if mac == hwaddr]
```
rosslagerwall, line 279: I think `list.extend()` is preferred.

upgrade.py
```diff
@@ -503,6 +531,24 @@ def completeUpgrade(self, mounts, prev_install, target_disk, backup_partnum, log
...
+        src_file_path = os.path.join(interface_rename_dir, 'dynamic-rules.json')
+        dest_file_path = os.path.join(
+            mounts['root'],
+            constants.FIRSTBOOT_DATA_DIR,
+            'initial_network_device_rules.conf'
+        )
+        if os.path.exists(src_file_path):
+            if self.convert_interface_rename_data(src_file_path, dest_file_path):
+                logger.log(f'Removing {interface_rename_dir} ...')
+                shutil.rmtree(interface_rename_dir)
```
rosslagerwall, line 547: Can't you read the data directly from the backup instead of copying it to the new installation and then immediately removing it?

upgrade.py
```diff
@@ -487,6 +487,34 @@ def buildRestoreList(self, src_base):
         return restore_list
 
     completeUpgradeArgs = ['mounts', 'installation-to-overwrite', 'primary-disk', 'backup-partnum', 'logs-partnum', 'net-admin-interface', 'net-admin-bridge', 'net-admin-configuration']
+
+    def convert_interface_rename_data(self, src_file_path, dest_file_path):
+        """
+        Convert data of interface-rename to the data consumed by networkd.
```
rosslagerwall, line 493: Since we use oth xcp-networkd and systemd-networkd you should be explicit here.

upgrade.py
```diff
@@ -487,6 +487,34 @@ def buildRestoreList(self, src_base):
...
+        else:
+            pattern = r'eth(\d+)'
+            lines = []
+            for (name, mac) in net_devs:
+                match = re.search(pattern, name)
+                if match:
+                    position = match.group(1)
+                    line = f"{position}:mac=\"{mac}\"\n"
+                    logger.log(f"Converted interface-rename data: {line}")
+                    lines += line
```
rosslagerwall, line 508: Does this actually work? `lines` is a list and `line` is a string and the result of adding them is unlikely to be what you want:

```
>>> x = []
>>> x += "hello"
>>> x
['h', 'e', 'l', 'l', 'o']
```

upgrade.py
```diff
@@ -487,6 +487,34 @@ def buildRestoreList(self, src_base):
...
+                else:
+                    logger.error(f'Network device name {name} is not like ethN.')
+                    return False
+
+            with open(dest_file_path, mode='w', encoding='utf-8') as f:
+                f.writelines(lines)
```
rosslagerwall, line 514: Instead of building up a list of lines and then writing them like this, maybe it would be better to just write the file line by line?

upgrade.py
```diff
@@ -487,6 +487,34 @@ def buildRestoreList(self, src_base):
...
+            for (name, mac) in net_devs:
+                match = re.search(pattern, name)
+                if match:
+                    position = match.group(1)
+                    line = f"{position}:mac=\"{mac}\"\n"
+                    logger.log(f"Converted interface-rename data: {line}")
+                    lines += line
+                else:
+                    logger.error(f'Network device name {name} is not like ethN.')
+                    return False
```
rosslagerwall, line 511: I don't think returning an error here is the right thing to do. If something failed to be named correctly during the previous boot, wouldn't it be better to write out the information we have and let the upgrade complete rather than failing it and leaving an unbootable system?

(Possibly for the above parse error too.)

rosslagerwall:
I've left some mostly minor comments, but I have a question at a higher level. It seems like the only thing it retrieves from the dynamic rules is the mapping between ethX number and MAC address. Doesn't XAPI/xcp-networkd already have this information in the database and it can then do the upgrade itself?

minglumlu:
> I've left some mostly minor comments, but I have a question at a higher level. It seems like the only thing it retrieves from the dynamic rules is the mapping between ethX number and MAC address. Doesn't XAPI/xcp-networkd already have this information in the database and it can then do the upgrade itself?

The mac-number mappings are in XAPI DB also, not in networkd db. But under this circumstance, xapi and networkd don't run. It has to parse from a file anyway, either xapi DB, or dynamic rules.  Parsing from dynamic rules seems simpler as there is existing code for it.

upgrade.py
minglumlu: OK. It seems this works (as I tested it as a whole upgrade) but not in expected way like this.
I can make it write line by line as you are suggesting in another comment.

rosslagerwall:
> > I've left some mostly minor comments, but I have a question at a higher level. It seems like the only thing it retrieves from the dynamic rules is the mapping between ethX number and MAC address. Doesn't XAPI/xcp-networkd already have this information in the database and it can then do the upgrade itself?
> 
> The mac-number mappings are in XAPI DB also, not in networkd db. But under this circumstance, xapi and networkd don't run. It has to parse from a file anyway, either xapi DB, or dynamic rules. Parsing from dynamic rules seems simpler as there is existing code for it.

I wasn't suggesting that the host-installer parses either DB.

The output of the code here is some config file which the toolstack is presumably going to parse on first boot. If the information is already available in a toolstack DB, why not let the toolstack use the information it already has on first boot and skip this intermediate step?

minglumlu:
> The output of the code here is some config file which the toolstack is presumably going to parse on first boot. If the information is already available in a toolstack DB, why not let the toolstack use the information it already has on first boot and skip this intermediate step?

The interface ordering shall be done by networkd which starts before the xapi. In other words, the xapi relies on the order established by the networkd. At least, the networkd can't depend on xapi.

upgrade.py
```diff
@@ -487,6 +524,7 @@ def buildRestoreList(self, src_base):
         return restore_list
 
     completeUpgradeArgs = ['mounts', 'installation-to-overwrite', 'primary-disk', 'backup-partnum', 'logs-partnum', 'net-admin-interface', 'net-admin-bridge', 'net-admin-configuration']
+
```
rosslagerwall, line 527: Not sure if this was intentional or accidental but none of the other functions have a blank line between `funcArgs = ...` and `def func(...`

upgrade.py
```diff
@@ -174,6 +174,44 @@ def restore_file(src_base, f, d=None, attr=None):
...
+            dest_file_path = os.path.join(
+                dest,
+                constants.FIRSTBOOT_DATA_DIR,
+                'initial_network_device_rules.conf'
+            )
+            net_devs = []
+            if os.path.exists(src_file_path):
+                net_devs = netutil.net_devs_of_last_boot(src_file_path)
+            else:
+                logger.log(f"The interface-rename data file {src_file_path} does not exist.")
```
rosslagerwall, line 197: Should the function return early here?

In future upgrades, there will never be any interface rename data so continuing to create the file below seems unnecessary.

upgrade.py
```diff
@@ -174,6 +174,44 @@ def restore_file(src_base, f, d=None, attr=None):
...
+
+            # The sorting result of interface-rename is to rename the interfaces to eth<N>
+            pattern = r'^eth(\d+)$'
```
rosslagerwall, line 200: For regexps used more than once, it would be better to do:

```
pattern = re.compile(r'^eth(\d+)$')
...
match = pattern.search(name)
```

upgrade.py
```diff
@@ -174,6 +174,44 @@ def restore_file(src_base, f, d=None, attr=None):
...
+            try:
+                with open(dest_file_path, mode='w', encoding='utf-8') as f:
+                    for (name, mac) in net_devs:
+                        match = re.search(pattern, name)
+                        if match:
+                            position = match.group(1)
+                            line = f"{position}:mac=\"{mac}\"\n"
```
rosslagerwall, line 207: You can simplify this slightly by using single quotes: `line = f'{position}:mac="{mac}"\n'`

upgrade.py
minglumlu: It's accidental :(

rosslagerwall (approved):

MarkSymsCtx:
Don't forget to squash the fixup commits

upgrade.py
```diff
@@ -174,6 +174,47 @@ def restore_file(src_base, f, d=None, attr=None):
...
+                constants.FIRSTBOOT_DATA_DIR,
+                'initial_network_device_rules.conf'
+            )
+            net_devs = []
+            if os.path.exists(src_file_path):
+                net_devs = netutil.net_devs_of_last_boot(src_file_path)
+            else:
+                logger.log(f"The interface-rename data file {src_file_path} does not exist.")
+
+            if not net_devs:
```
liulinC, line 199: you can just return in above else, but not a big deal.

liulinC (approved):

upgrade.py
```diff
@@ -174,6 +174,47 @@ def restore_file(src_base, f, d=None, attr=None):
...
+                return
+
+            # The sorting result of interface-rename is to rename the interfaces to eth<N>
+            pattern = re.compile(r'^eth(\d+)$')
+            try:
+                with open(dest_file_path, mode='w', encoding='utf-8') as f:
```
liulinC, line 205: Hope to use this in some days 😄 
```
net_devs 
|> List.filter_map(fun name, mac ->  match pattern.match | xxx-> Some xxx | None)
|> List.iter( fun mac -> f.write(mac))
```

PR #253, opened 2025-05-27
Support DMV function in host-installer

chunjiez:
Please refer to the internal design doc on the Feature Home page for the visual effect.

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1574,15 +1574,7 @@ def writeDMVSelections(mounts, selected_multiversion_drivers):
         # if we got empty selection we need more log data to see devices and
         # drivers that we have
         drivers = dmv_data_provider.getDriversData()
-        for d in drivers:
-            logger.log("driver: %s" % d.getHumanDriverLabel())
-            logger.log("device list:")
-            for l in d.getHumanDeviceLabel():
-                logger.log(l)
-            logger.log("variants:")
-            for v in d.variants:
-                logger.log(v)
-            logger.log("")
+        dmvutil.logDriverVariants(drivers)
 
         selected_multiversion_drivers = dmv_data_provider.chooseDefaultDriverVariants()
         logger.log("pass default driver variants to driver-tool.")
--- a/dmvutil.py
+++ b/dmvutil.py
@@ -151,7 +151,6 @@ def getDeviceDriverMap(pci_dev_list):
             l = device_driver_map[driver]
             if pci_dev.getHumanDevLabel() not in l:
                 l.append(pci_dev.getHumanDevLabel())
-            device_driver_map[driver] = l
         else:
             device_driver_map[driver] = [pci_dev.getHumanDevLabel()]
     return device_driver_map
@@ -314,7 +313,7 @@ class DriverMultiVersionData:
             return False
         return True
 
-    def selectMultiDriverVariants(self, choices):
+    def applyDriverVariants(self, choices):
         failures = []
         for driver_name, variant_name in choices:
             ret = self.selectSingleDriverVariant(driver_name, variant_name)
@@ -343,3 +342,14 @@ def getDMVData():
         raise RuntimeError("Failed to execute 'lspci'")
     logger.log(devlist)
     return DriverMultiVersionData(dmvlist, devlist)
+
+def logDriverVariants(drivers):
+    for d in drivers:
+        logger.log("driver: %s" % d.getHumanDriverLabel())
+        logger.log("device list:")
+        for l in d.getHumanDeviceLabel():
+            logger.log(l)
+        logger.log("variants:")
+        for v in d.variants:
+            logger.log(v)
+        logger.log("")
--- a/tui/installer/screens.py
+++ b/tui/installer/screens.py
@@ -491,21 +491,13 @@ def dmv_screen(answers):
 
     drivers = []
     hw_present_drivers = []
-    if not "selected-multiversion-drivers" in answers:
+    if "selected-multiversion-drivers" not in answers:
         answers['selected-multiversion-drivers'] = []
 
     if not dmv_data_provider:
         dmv_data_provider = dmvutil.getDMVData()
         drivers = dmv_data_provider.getDriversData()
-        for d in drivers:
-            logger.log("driver: %s" % d.getHumanDriverLabel())
-            logger.log("device list:")
-            for l in d.getHumanDeviceLabel():
-                logger.log(l)
-            logger.log("variants:")
-            for v in d.variants:
-                logger.log(v)
-            logger.log("")
+        dmvutil.logDriverVariants(drivers)
 
     # skip the ui rendering
     hw_present_drivers = dmv_data_provider.getHardwarePresentDrivers()
@@ -611,12 +603,18 @@ def confirm_dmv_selection(answers):
 
             for drvname, oemtype in choices:
                 logger.log("select and enable variant %s for driver %s." % (oemtype, drvname))
-            failures = dmv_data_provider.selectMultiDriverVariants(choices)
+            failures = dmv_data_provider.applyDriverVariants(choices)
             if len(failures) == 0:
                 logger.log("succeed to select and enable all driver variants.")
             else:
                 for driver_name, variant_name in failures:
-                    logger.log("fail to select and enable variant %s for driver %s." % (variant_name, driver_name))
+                    logger.log("fail to select or enable variant %s for driver %s." % (variant_name, driver_name))
+                ButtonChoiceWindow(
+                        tui.screen,
+                        "Problem Loading Driver Variant",
+                        "Setup was unable to activate driver variant.",
+                        ['Ok']
+                        )
             return RIGHT_FORWARDS
         else:
             title = "Error"


chunjiez:
DMV, stands for Driver Multi Version. DMV provides capability to distribute new OEM driver versions (which do not regress functionality) alongside other host updates.

- latest driver updates are installed along with other host updates
- reduces engineering overhead of creating separate driver disks
- if kABI changes occur then it is simpler for engineering to rebuild and reissue all associated driver RPMs plus the kABI associated kernel change.

The code change in XSConsole provides a mechanism for selecting one variant of a multi-version driver.

psafont:
There's more public documentation about this feature on the xapi project: https://github.com/xapi-project/xen-api/blob/d27a1181bbf043ed00eb0b2d61fe434d8588d2d2/doc/content/toolstack/features/MVD/index.md

dmvutil.py
```diff
@@ -0,0 +1,253 @@
...
+def queryDriversOrVariant(context):
+    if isinstance(context, Drivers):
+        return ("drivers", context)
+    elif isinstance(context, list):
+        return ("variants", queryMultipleVariants(context))
+    elif isinstance(context, DriverVariant):
+        return ("variant", querySingleVariant(context))
+    return ("unknown", None)
+
+def sameDriverMultiVariantsSelected(variants):
```
rosslagerwall, line 238: This entire function seems quite inelegant. I think you can use `itertools.groupby()` to do it in a nicer way.

dmvutil.py
```diff
@@ -0,0 +1,253 @@
...
+        self.drvname = drvname
+        self.type = ""
+        self.friendly_name = ""
+        self.description = ""
+        self.info = ""
+        self.selected = None
+        self.active = None
+        self.variants = []
+        self.devices = []
+        if drvinfo != None:
```
rosslagerwall, line 65: `if drvinfo:` is simpler.

dmvutil.py
```diff
@@ -0,0 +1,253 @@
...
+            driver.setDeviceList(d.devices)
+            if len(driver.variants) > 0:
+                hw_present_drivers = driver
+    return hw_present_drivers
+
+def chooseDefaultDriverVariants(drivers):
+    variants = []
+    for d in drivers:
+        v = d.selectVariantByDefault()
+        if v != None:
```
rosslagerwall, line 219: `if v:` is simpler.

dmvutil.py
```diff
@@ -0,0 +1,253 @@
...
+    newdrv.friendly_name = olddrv.friendly_name
+    newdrv.description = olddrv.description
+    newdrv.info = olddrv.info
+    newdrv.selected = olddrv.selected
+    newdrv.active = olddrv.active
+
+def getHardwarePresentDrivers(drivers):
+    hw_present_drivers = []
+    for d in drivers:
+        driver = Drivers(d.drvname)
```
rosslagerwall, line 195: Why not change the `Drivers()` constructor so that it can take a list of variants and pass the filtered list below?

That way you don't have to construct a Drivers object with variants, copy it and throw the variants away, then set a new list of variants.

(Would there be any variants to throw away anyway since you haven't passed in driver info?)

dmvutil.py
```diff
@@ -0,0 +1,253 @@
...
+            variants.append(v)
+    return variants
+
+def queryMultipleVariants(context):
+    return context
+
+def querySingleVariant(context):
+    return context
+
+def queryDriversOrVariant(context):
```
rosslagerwall, line 229: This function seems like a code smell to me. Why does the caller not know what object it is dealing with?

dmvdata.py
```diff
@@ -0,0 +1,186 @@
...
+
+def getRealDMVList():
+    rc, out = util.runCmd2(['driver-tool', '-l'], with_stdout=True)
+    if rc != 0:
+        logger.log(out)
+        return None
+    return out
+
+def getRealHardwareList():
+    rc, out = util.runCmd2(['lspci'], with_stdout=True)
```
rosslagerwall, line 172: I think you should use `-mm` to get machine readable output (otherwise the format could change).

Also pass `-D` so that `lspci` always outputs the domain then you scan skip the `pci_id = "0000:" + pci_slot` hack.

dmvdata.py
```diff
@@ -0,0 +1,186 @@
...
+    if rc != 0:
+        logger.log(out)
```
rosslagerwall, line 174: No need to explicitly log the output since runCmd2 does it anyway.

dmvdata.py
```diff
@@ -0,0 +1,186 @@
...
+    def sameDriverMultiVariantsSelected(self, variants):
+        return dmvutil.sameDriverMultiVariantsSelected(variants)
+
+    def chooseDefaultDriverVariants(self, drivers = None):
+        if not drivers:
+            drivers = self.hw_present_drivers
+        return dmvutil.chooseDefaultDriverVariants(drivers)
+
+    def selectSingleDriverVariant(self, driver_name, variant_name):
+        cmdstring = "driver-tool|-s|-n|%s|-v|%s" % (driver_name, variant_name)
```
rosslagerwall, line 52: Same comment as above about argument list vs strings.

dmvdata.py
```diff
@@ -0,0 +1,186 @@
...
+        rc, out = util.runCmd2(cmdstring.split('|'), with_stdout=True)
+        if rc != 0:
+            logger.log(out)
+            return False
+
+        rc, out = util.runCmd2(['rmmod', driver_name], with_stdout=True)
```
rosslagerwall, line 58: I don't think this would work reliably.

If the driver is composed of multiple modules this will only rmmod one of them and then the modprobe could end of with a mix of modules.
If the driver is in use (e.g. because a RAID device was automatically assembled from the devices it provides), then rmmod will fail and the modprobe below will fail too.

dmvdata.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def selectMultiDriverVariants(self, choices):
+        failures = []
+        for driver_name, variant_name in choices:
+            ret = self.selectSingleDriverVariant(driver_name, variant_name)
+            if not ret:
+                failures.append((driver_name, variant_name))
+        return failures
+
+def getMockHardwareList():
```
rosslagerwall, line 77: I don't think all this mock data should live in the code itself. It should go in `dmvtest.py`.

Furthermore, the other functions should not have `Real` in the name.

dmvdata.py
```diff
@@ -0,0 +1,186 @@
...
+
+# SPDX-License-Identifier: GPL-2.0-only
+
+import json
+from json.decoder import JSONDecodeError
+import util
+import dmvutil
+from xcp import logger
+
+class DriverMultiVersionData:
```
rosslagerwall, line 11: Maybe all of this (after the mock data is removed) could live in `dmvutil.py`? It's not really clear to me why this separation exists.

dmvutil.py
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env python3
```
rosslagerwall, line 1: It would be nice if this dmv library code could be shared between the installer and other uses (e.g. driver tool and xsconsole). There isn't really anything installer-specific about most of it.

tui/installer/screens.py
```diff
@@ -449,6 +451,190 @@ def backup_existing_installation(answers):
...
+
+            if button is None or button == 'back': return LEFT_BACKWARDS
+
+            for drvname, oemtype in choices:
+                logger.log("select and enable variant %s for driver %s." % (oemtype, drvname))
+            failures = dmv_data_provider.selectMultiDriverVariants(choices)
+            if len(failures) == 0:
+                logger.log("succeed to select and enable all driver variants.")
+            else:
+                for driver_name, variant_name in failures:
```
stephenchengCloud, line 625: Seems `driver_name` and `variant_name` are not used.
And `item` is not defined in the log scope.

backend.py
```diff
@@ -1563,6 +1564,16 @@ def touchSshAuthorizedKeys(mounts):
     fh = open("%s/root/.ssh/authorized_keys" % mounts['root'], 'a')
     fh.close()
 
+def writeDMVSelections(mounts, selected_multiversion_drivers):
+    for driver_name, variant_name in selected_multiversion_drivers:
+        logger.log("write variant %s selection for driver %s." % (variant_name, driver_name))
```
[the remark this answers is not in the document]
chunjiez: Here I simply follow the old-style to do the log output.

If I use f-strings style, then the style of new code and old code looks inconsistent.

backend.py
```diff
@@ -1563,6 +1564,16 @@ def touchSshAuthorizedKeys(mounts):
...
+
+        cmdstring = "driver-tool|-s|-n|%s|-v|%s" % (driver_name, variant_name)
```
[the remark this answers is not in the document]
chunjiez: Will update.

backend.py
```diff
@@ -1563,6 +1564,16 @@ def touchSshAuthorizedKeys(mounts):
...
+        chrootcmd = ['chroot', mounts['root']]
+        chrootcmd.extend(cmdstring.split('|'))
+        rc, out = util.runCmd2(chrootcmd, with_stdout=True)
+        if rc != 0:
+            logger.log(out)
```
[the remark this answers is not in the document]
chunjiez: Will update.

Inside runCmd2, it already catches any possible exception.

tui/installer/screens.py
```diff
@@ -449,6 +451,190 @@ def backup_existing_installation(answers):
...
+    for drvname, oemtype in choices:
+        v = dmv_data_provider.getDriverVariantByName(drvname, oemtype)
+        if v:
+            variants.append(v)
+
+    entries = []
+    title = "Confirm Drivers Selection"
+    text = ""
+    if len(variants) == 0:
+        #text = "No drivers selected? You can select drivers when your system comes online after host installation.\n\nContinue or go back to driver selection"
```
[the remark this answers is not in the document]
chunjiez: Will update.

dmvutil.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def __repr__(self):
+        return "<drivers: %s (%s)>" % (self.friendly_name, self.description)
+
+    def setDeviceList(self, devices):
+        self.devices = devices
+
+    def getHumanDriverLabel(self):
+        template = "{friendly_name} ({info})"
+        return template.format(friendly_name=self.friendly_name, info=self.info)
```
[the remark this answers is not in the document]
chunjiez: Will update.

dmvutil.py
```diff
@@ -0,0 +1,253 @@
...
+        variants = []
+        template = "{description} variant: {vartype}"
+        for variant in self.variants:
+            label = template.format(description=self.description, vartype=variant.oemtype)
+            variants.append((label, variant))
+        return variants
+
+    def selectVariantByDefault(self):
+        variant = None
+        l = [float(v.priority) for v in self.variants]
```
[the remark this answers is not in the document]
chunjiez: in case of some day priority is assigned with a float number

dmvutil.py
```diff
@@ -0,0 +1,253 @@
...
+    for d in drivers:
+        driver = Drivers(d.drvname)
+        cloneDriverWithoutVariants(d, driver)
+        driver.variants = list(filter(lambda x:x.hardware_present, d.variants))
+        driver.setDeviceList(d.devices)
+        if len(driver.variants) > 0:
+            hw_present_drivers.append(driver)
+    return hw_present_drivers
+
+def getHardwarePresentDriver(drivers, name):
```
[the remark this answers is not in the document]
chunjiez: Will update.

dmvutil.py
chunjiez: Will update.

dmvutil.py
chunjiez: Will update.

dmvutil.py
chunjiez: In DMV screen, it might render driver or variant details page if users want to see the details, main screen passes driver or variant object to the details page, so, the details page need to know the object type, driver or variant, then it can render different details.

chunjiez: In DMV screen, it might render driver or variant details page if users want to see the details, main screen passes driver or variant object to the details page, so, the details page need to know the object type, driver or variant, then it can render different details.

dmvdata.py
chunjiez: Will update.

dmvdata.py
chunjiez: Will update.

dmvdata.py
chunjiez: Will update.

dmvutil.py
```diff
@@ -0,0 +1,345 @@
...
+
+def getDeviceDriverMap(pci_dev_list):
+    device_driver_map = {}
+    for pci_dev in pci_dev_list:
+        driver = pci_dev.driver
+        if driver in device_driver_map:
+            l = device_driver_map[driver]
+            if pci_dev.getHumanDevLabel() not in l:
+                l.append(pci_dev.getHumanDevLabel())
+            device_driver_map[driver] = l
```
DeliZhangX, line 154: This line is needless.
Also you can use `set` which will avoid duplicated elements.

tui/installer/screens.py
```diff
@@ -449,6 +451,183 @@ def backup_existing_installation(answers):
...
+    return True
+
+# driver multi version screen:
+def dmv_screen(answers):
+    global dmv_data_provider
+
+    drivers = []
+    hw_present_drivers = []
+    if not "selected-multiversion-drivers" in answers:
+        answers['selected-multiversion-drivers'] = []
```
DeliZhangX, line 495: Use `setdefault()`

backend.py
```diff
@@ -1563,6 +1565,35 @@ def touchSshAuthorizedKeys(mounts):
...
+        drivers = dmv_data_provider.getDriversData()
+        for d in drivers:
+            logger.log("driver: %s" % d.getHumanDriverLabel())
+            logger.log("device list:")
+            for l in d.getHumanDeviceLabel():
+                logger.log(l)
+            logger.log("variants:")
+            for v in d.variants:
+                logger.log(v)
+            logger.log("")
```
DeliZhangX, line 1585: Could let these logging as a function, and call here and in `dmv_screen()`

backend.py
```diff
@@ -1563,6 +1565,35 @@ def touchSshAuthorizedKeys(mounts):
...
+
+        selected_multiversion_drivers = dmv_data_provider.chooseDefaultDriverVariants()
```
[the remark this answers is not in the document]
DeliZhangX: Just ignore this comment if do not need selected_multiversion_drivers to return info.

backend.py
chunjiez: Will update.

tui/installer/screens.py
```diff
@@ -449,6 +451,183 @@ def backup_existing_installation(answers):
...
+                tui.screen, title, text,
+                entries,
+                ['Ok', 'Back'], 60, scroll, height, None, help='dmv:info3',
+                hotkeys={})
+
+            if button is None or button == 'back': return LEFT_BACKWARDS
+
+            for drvname, oemtype in choices:
+                logger.log("select and enable variant %s for driver %s." % (oemtype, drvname))
+            failures = dmv_data_provider.selectMultiDriverVariants(choices)
```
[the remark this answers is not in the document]
chunjiez: Will update selectMultiDriverVariants to applyDriverVariants

"Also if apply failed, do we should remind customer to reselect?" This is useful on XenCenter. The host installer UI is not same as Desktop or Web UI. Here if driver selection fails, then most likely, host installation would be broken, users would seek support engineers' help.

tui/installer/screens.py
chunjiez: We have many "xxxxxx not in answers" python code instances.

I prefer to stick to the old style.

dmvutil.py
chunjiez: Will update.

tui/installer/screens.py
chunjiez: code updated.

DeliZhangX (approved):

rosslagerwall (approved):

PR #254, opened 2025-05-29
CP-54461: SWRAID Upgrade & Restore

GeraldEV:
Dependent on PR https://github.com/xenserver/host-installer/pull/248

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1243,10 +1243,6 @@ def addEfiBootEntry(mounts, disk, boot_partnum, install_type, branding):
                             "-d", disk, "-p", str(boot_partnum)], with_stderr=True)
     check_efibootmgr_err(rc, err, install_type, "Failed to add new efi boot entry")
 
-def setEfiBootEntry(mounts, disk, boot_partnum, install_type, branding):
-    clearEfiBootEntries(mounts, install_type, branding)
-    addEfiBootEntry(mounts, disk, boot_partnum, install_type, branding)
-
 def writeBootEntries(write_boot_entry, mounts, disks, primary_partnum, boot_partnum, target_boot_mode, install_type, branding, location):
     if target_boot_mode == TARGET_BOOT_MODE_UEFI and write_boot_entry:
         clearEfiBootEntries(mounts, install_type, branding)
--- a/product.py
+++ b/product.py
@@ -583,10 +583,9 @@ def findXenSourceProducts():
     installs = []
 
     for disk_device in diskutil.getQualifiedDiskList():
-        disk = diskutil.probeDisk(disk_device)
-
         inst = None
         try:
+            disk = diskutil.probeDisk(disk_device)
             if disk.root[0] == diskutil.INSTALL_RETAIL:
                 inst = ExistingRetailInstallation(disk_device, disk.boot[1], disk.root[1], disk.state[1], disk.storage)
         except Exception as e:
--- a/restore.py
+++ b/restore.py
@@ -151,14 +151,12 @@ def restoreFromBackup(backup, progress=lambda x: ()):
 
         # mount restore partition:
         dest_fs = util.TempMount(restore_partition, 'restore-dest-')
-        efi_mounted = False
         try:
-            mounts = {'root': dest_fs.mount_point, 'boot': os.path.join(dest_fs.mount_point, 'boot')}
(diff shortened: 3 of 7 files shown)


backend.py
```diff
@@ -1717,7 +1713,7 @@ def isSWRAIDSyncd(primary_disk):
         logger.log("Failed to check if SWRAID device is sync'd due to: " + str(ex))
         return False
 
-def completeSWRAID(mounts, primary_disk, physical_disks, write_boot_entry, install_type, branding, boot_partnum, primary_partnum, logs_partnum, target_boot_mode):
+def completeSWRAID(mounts, primary_disk, physical_disks, write_boot_entry, install_type, branding, boot_partnum, primary_partnum, logs_partnum, target_boot_mode, remount=True):
```
rosslagerwall, line 1716: Having a boolean parameter to call `mountVolumes()` like this doesn't feel right. Why not just add `mountVolumes()` to the finalization sequence?

(Though perhaps some of this might change based on the discussion in #248.)

rosslagerwall (commented):
I've reviewed the top commit only since the others were reviewed in #248.

product.py
```diff
@@ -583,10 +583,9 @@ def findXenSourceProducts():
     installs = []
 
     for disk_device in diskutil.getQualifiedDiskList():
-        disk = diskutil.probeDisk(disk_device)
-
         inst = None
         try:
+            disk = diskutil.probeDisk(disk_device)
```
rosslagerwall, line 588: Why is this change needed? Is there a new reason why `probeDisk()` can fail?

tui/installer/screens.py
```diff
@@ -326,6 +326,9 @@ def more_info(context):
             answers['primary-disk'] = answers['installation-to-overwrite'].primary_disk
             answers['physical-disks'] = [answers['primary-disk']]
 
+            if diskutil.is_raid(entry):
+                answers['physical-disks'] = diskutil.getSWRAIDDevices(entry)
```
rosslagerwall, line 330: Maybe line 327 should be in an `else` clause here?

tui/installer/screens.py
```diff
@@ -617,6 +620,8 @@ def select_primary_disk(answers):
     # entry contains the 'de' part of the tuple passed in
     answers['primary-disk'] = entry
     answers['physical-disks'] = [entry]
+    if diskutil.is_raid(entry):
+        answers['physical-disks'] = diskutil.getSWRAIDDevices(entry)
```
rosslagerwall, line 624: ... and same here

restore.py
```diff
@@ -203,29 +203,49 @@ def restore_partitions():
                 (tool, restore_partition) = restore_partitions()
                 dest_fs = util.TempMount(restore_partition, 'restore-dest-')
 
+            if diskutil.is_raid(disk_device):
+                backend.waitForSWRAIDSync(disk_device)
+
+                # umount and re-mount partitions to flush the cache
+                # umount all partitions first before re-mounting
+                if efi_mounted and "esp" in mounts:
```
rosslagerwall, line 211: Rather than adding more `efi_mounted` conditionals, I think it would be better to add a commit before this that cleans up this function and drops all the conditional EFI stuff from this function. XS9 only supports upgrade from UEFI systems, therefore restores will always be back to a UEFI system.

restore.py
```diff
@@ -213,7 +213,12 @@ def restore_partitions():
                 if efi_boot:
                     branding = dict(inventory)
                     branding['product-brand'] = branding['PRODUCT_BRAND']
-                    backend.setEfiBootEntry(mounts, disk_device, boot_partnum, constants.INSTALL_TYPE_RESTORE, branding)
+                    backend.clearEfiBootEntries(mounts, constants.INSTALL_TYPE_RESTORE, branding)
+                    if disk_device.startswith('/dev/md'):  # Using SWRAID
+                        physical_disks = util.getSWRAIDDevices(disk_device)
```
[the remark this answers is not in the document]
GeraldEV: It's possible at this point that one of the disks is unavailable - so while we would like there to be two disks present, it might not be the case

product.py
rosslagerwall: OK, makes sense. There are multiple users of `probeDisk()`. Do the other users need to be adjusted too?

PR #255, opened 2025-05-29
CA-410594: Settle udev after eject

rosslagerwall:
Calling eject on a CD-ROM causes some udev rules to run. In some cases, the rules will poke /dev and race with the unmount of /dev in writeMachineId() and cause the unmount to fail with EBUSY. The failing unmount is ignored which ends up with /dev being mounted on /tmp/root twice causing the final unmount of /tmp/root to fail (because one instance of /dev is still bind-mounted). Fix this by settling udev after eject() so the system is in a clean state before continuing.

### DIFF
--- a/repository.py
+++ b/repository.py
@@ -12,6 +12,7 @@ import subprocess
 import re
 import gzip
 import shutil
+import time
 from io import BytesIO
 from xml.dom.minidom import parse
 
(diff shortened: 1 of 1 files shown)


freddy77:
Should this be backported to xs8 too?

PR #260, opened 2025-06-13
CA-400058: Fail if there is no first partition to preserve when requested

GeraldEV:
The previous PR (https://github.com/xenserver/host-installer/pull/198/files) for this improvement assumed that the first partitions are always utility partitions and was hence reverted after internal tests failed

There's no specification or requirement that the first partition is a utility partition, but there is an option (the default) to preserve the first partition if it is a utility partition. This was likely the cause of the confusion

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -558,6 +558,8 @@ def partitionTargetDisk(disk, existing, preserve_first_partition, create_sr_part
     # Preserve any utility partitions unless user told us to zap 'em
     primary_part = 1
     if preserve_first_partition == 'true':
+        if tool.getPartition(1):  # If no first partition
+            raise RuntimeError("No first partition to preserve")
         primary_part += 1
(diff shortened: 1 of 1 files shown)


chunjiez (dismissed):

backend.py
```diff
@@ -558,6 +558,8 @@ def partitionTargetDisk(disk, existing, preserve_first_partition, create_sr_part
     # Preserve any utility partitions unless user told us to zap 'em
     primary_part = 1
     if preserve_first_partition == 'true':
+        if 1 not in tool.partitionTable():  # If no first partition
```
rosslagerwall, line 561: This method triggers the partition tool to rescan the disk which it has already done in the constructor. Generally this method should not be used publicly.

I think you want something like `if tool.getPartition(1):` instead.

freddy77 (approved):

PR #261, opened 2025-06-23
CA-412051: Add debug logs for libsystemd-core.so link issue

DeliZhangX:
After installation, the following error occurred during startup.

  /init: error while loading shared libraries:
    libsystemd-core-257.2-5.xs9.so: cannot open shared object file:
    No such file or directory

This problem is extremely difficult to reproduce, so added debug logs, and which will be reverted in future.

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -854,10 +854,15 @@ def __mkinitrd(mounts, primary_disk, partition, kernel_version):
     except:
         pass
 
-    cmd = ['dracut', '-f', output_file, kernel_version]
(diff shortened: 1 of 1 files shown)


rosslagerwall:
While this is fine, I think a better approach may be to have XenRT automatically pause-of-fail when it encounters this issue. You can then boot into the installer shell and inspect the root disk to find out how it differs from a working installation.

stephenchengCloud (approved):

PR #238, opened 2025-04-08
CP-53840: Do not write mgmt interface info in xensource-inventory

liulinC:
The management interface information is maintained by networkd host-installer just give up writing this information and networkd will maintain and write that information

PR #262, opened 2025-06-23
CP-307931 Remove unused net-admin-interface key

changlei-li:
There is hard code "eth" in converting interface name to bridge name. It is incorrect after dropping interface-rename. As host-installer needn't write management-interface to inventory any more (see https://github.com/xenserver/host-installer/pull/238), the net-admin-bridge generating code and related parameters can be removed.


changlei-li:
Test manually upgrade from 8.4 to 9 OK.

DeliZhangX (approved):

PR #267, opened 2025-07-02
CA-413008: fix variant selection bug

chunjiez:
We should check variant selection for each driver in case of users' mistake operations which deselect variant for some drivers.

截屏2025-07-02 15 54 27


freddy77:
You should describe in the commit message what kind of bug you are fixing.

PR #271, opened 2025-07-11
Give a more explicit message when DNF gets killed

ydirson:


freddy77 (approved):

repository.py
```diff
@@ -850,7 +850,10 @@ def updateInstallProgress(m):
         rv = p.wait()
 
         if rv:
-            logger.log("DNF exited with %d" % rv)
+            if rv > 0:
+                logger.log("DNF exited with %d" % rv)
+            else:
+                logger.log("DNF killed with signal %d" % -rv)
```
rosslagerwall, line 856: Is it worth decoding the signal name too?

```
logger.log("DNF killed with signal: %d" % (signal.strsignal(-rv))
```

freddy77: and `%s` instead of `%d`.

rosslagerwall (approved):

freddy77 (approved):

PR #272, opened 2025-07-11
Give a more explicit message when Yum gets killed

ydirson:
This is a backport for xs8 of #271 but without translating signal number to name, since that function comes with python3


freddy77 (dismissed):

freddy77:
I suppose this now needs updating (original merged change was updated)

freddy77 (approved):

PR #277, opened 2025-07-21
CP-53110: Remove support for loading driver disks

stephenchengCloud:
- Dropped the ability to load additional driver disks.
- Removed FCoE configuration from the welcome screen. (Full FCoE removal will be completed in a follow-up.)

These features are no longer supported.

Tested:
Before:
image
After:
image


ydirson:
Could it be considered to move driver disks behind a feature flag instead?  Even with DMV they still have value: if their support is dropped, it can be good to keep the ability to use a newer driver not included in the latest ISO...

tui/installer/screens.py
```diff
@@ -45,65 +44,16 @@ def selectDefault(key, entries):
...
-            tui.update_help_line([None, "<F9> load driver"])
-        elif popup == 'storage':
-            tui.fcoe.select_fcoe_ifaces(answers)
-            tui.update_help_line([None, "<F9> load driver"])
-
-    tui.screen.popHelpLine()
+                                ['Ok', 'Cancel Installation'], width = 60)
 
-    if button == 'reboot':
+    if button == 'cancel installation':
```
chunjiez, line 56: After returning EXIT, the whole system would reboot. So, personally, I prefer the old string.

I suggest to contact with documentation team, see which string is better.

stephenchengCloud:
> > > > Could it be considered to move driver disks behind a feature flag instead? Even with DMV they still have value: if their support is dropped, it can be good to keep the ability to use a newer driver not included in the latest ISO...
> > > 
> > > 
> > > The ability to install supplemental packs remains, so we can still install drivers not included in the latest ISO. @rosslagerwall
> > 
> > 
> > F9 is to install driver in installer env; Supplemental packs install is to dom0 env.
> 
> Yes, the problem happens when eg. a user needs a driver for a RAID controller for the system disk, the driver is needed in the installer env first.

Thanks for your comments. We decided to keep driver disks and hide it behind a feature flag.

ydirson:
> Thanks for your comments. We decided to keep driver disks and hide it behind a feature flag.

Thanks!

PR #278, opened 2025-07-22
Fix console getting confused by unicode characters

ydirson:
The OS should have set a proper environment for the console to draw
line characters.  Disabling UTF-8 just breaks this when the console
has been properly configured for UTF8.

### DIFF
--- a/init
+++ b/init
@@ -163,9 +163,10 @@ def main(args):
 
     # start the user interface:
     if ui:
-        # switch to ISO 8859-1 mode so line drawing characters work as expected on
-        # vt100 terminals.
-        print("\033%@")
+        # switch to UTF-8 mode for line drawing characters, to reflect
+        # NEWT's behaviour of forcefully outputint UTF-8 regardless of
+        # the locale
+        print("\033%G")
 
         logger.log("Starting 'init' user interface on %s" % tty)
         ui.init_ui()


freddy77:
I didn't find that ISO 8859-1 setting anywhere (`ESC %@`). But it looks wrong anyway. First, ISO 8859-1 has not the semi graphics characters we need (like borders). Second today the world is more around using utf-8 not iso8859-1 (although I doubt vt100 supports utf-8 in all versions).
When I test with Qemu it looks like the borders are fine, when connected using Putty they looks wrong.
Which environment are you finding this issue?

bernhardkaindl (commented):
> I didn't find that ISO 8859-1 setting anywhere (ESC %@). 

https://linux.die.net/man/7/utf-8 (From the Linux `manpages` package):

> The official ESC sequence to switch from an ISO 2022 encoding scheme (as used for instance by VT100 terminals) to UTF-8 is ESC % G ("\x1b%G"). The corresponding return sequence from UTF-8 to ISO 2022 is ESC % @ ("\x1b%@"). Other ISO 2022 sequences (such as for switching the G0 and G1 sets) are not applicable in UTF-8 mode. 

> When I test with Qemu it looks like the borders are fine, when connected using Putty they looks wrong.

Putty has a config:
Window → Translation → "Enable VT100 line drawing even in UTF-8 mode".
and it's not enabled by default.

More details on how to configure Putty versions behave on Linux/Windows: https://stromasys.atlassian.net/wiki/spaces/KBP/pages/15138858/Cannot+get+correct+line+drawing+on+Linux+UNIX+with+Putty

So sending "\x1b%@" to putty should disable UTF-8 mode and allow the line drawing to work in Putty. That sounds messed up.

@ydirson I'm guessing that this is for putty:
To keep the sequence for putty, the code could check if the tty is a local Linux console tty or serial / pseudo tty and not emit the sequence for real Linux console ttys?

bernhardkaindl (commented):
I want to send this idea:

But as this image shows, the default radiobutton selector for "Handling of line drawing characters" is: "Use Unicode line drawing code points":


As https://linux.die.net/man/7/utf-8 says:
> The official ESC sequence to switch from an ISO 2022 encoding scheme (as used for instance by VT100 terminals) to UTF-8 is ESC % G ("\x1b%G"). 

Thus, if we use 
```py
print("\033%G") 
```
to instead switch all terminals (including putty) to UTF-8 mode then it should work for new Linux consoles and Putty, I'd hope.

ydirson:
This tty mode change naturally does not impact just host-installer, so for diag purposes `pstree` is a great way to quickly check the effects of such changes.
What I get is
* out of the box, `pstree` shows semigraphics as expected
* after `printf "\033%@"` the `pstree` output (like that of the installer TUI) uses 1 character per byte in the UTF8-encoded semigraphics, completely messing up the display

PR #284, opened 2025-07-24
CP-308419: Port SW RAID changes to master

GeraldEV:
Series of changes reviewed as part of the merge to release/xs8
Some patches had to be updated to be compatible with the upstream changes since the earlier release.

During testing found and fixed the following issues:
Fix restore partnum: https://github.com/xenserver/host-installer/pull/281
Fix first partition check: https://github.com/xenserver/host-installer/pull/282

Tested manual and answerfile install/upgrade/restore with and without SW RAID

### DIFF
--- a/restore.py
+++ b/restore.py
@@ -92,7 +92,7 @@ def restoreFromBackup(backup, progress=lambda x: ()):
         dest_fs = util.TempMount(restore_partition, 'restore-dest-')
         try:
             mounts = {'root': dest_fs.mount_point,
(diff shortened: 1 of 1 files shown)


freddy77 (approved):

GeraldEV:
> Does this port mean this RAID support will be maintained after all?

The support statement from the previous PRs has not changed, only the internal customer which requested the feature
It is being forward-ported to master to keep minimise differences between the releases and to enable the customer to upgrade in the future with extra implementation effort
We will not be supporting SW RAID for the general customer base

PR #286, opened 2025-07-28
CP-308812 Add new Answerfile label ssh-mode to control SSH status

LunfanZhang:
Add a new answerfile label `<ssh-mode>on/off/auto</ssh-mode>` to control SSH auto-mode behavior and SSH service status:

- `<ssh-mode>on</ssh-mode>`: Disables auto-mode and enables SSH by default
- `<ssh-mode>off</ssh-mode>`: Disables auto-mode and disables SSH by default
- `<ssh-mode>auto</ssh-mode>`: Enables auto-mode and stops SSH once XAPI starts up

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -182,6 +182,7 @@ def getFinalisationSequence(ans):
         Task(prepareSwapfile, A(ans, 'mounts', 'primary-disk', 'swap-partnum', 'disk-label-suffix'), []),
         Task(writeFstab, A(ans, 'mounts', 'primary-disk', 'logs-partnum', 'swap-partnum', 'disk-label-suffix', 'fs-type'), []),
         Task(enableAgent, A(ans, 'mounts', 'network-backend', 'services'), []),
+        Task(configureSSHMode, A(ans, 'mounts', 'ssh-mode'), []),
         Task(configureCC, A(ans, 'mounts'), []),
         Task(writeInventory, A(ans, 'installation-uuid', 'control-domain-uuid', 'mounts', 'primary-disk',
                                'backup-partnum', 'logs-partnum', 'boot-partnum', 'swap-partnum', 'storage-partnum',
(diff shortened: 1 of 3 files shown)


freddy77 (dismissed):

stephenchengCloud (dismissed):

backend.py
```diff
@@ -1280,6 +1281,36 @@ def writeFstab(mounts, primary_disk, logs_partnum, swap_partnum, disk_label_suff
...
 
+def configureSSHMode(mounts, ssh_mode):
+    """Configure SSH mode according to answerfile settings."""
+
+    if ssh_mode is None or ssh_mode not in ['auto', 'on', 'off']:
+        return
+
+    logger.log("Configuring SSH mode: %s" % ssh_mode)
+
+    ssh_auto_mode_conf_path = os.path.join(mounts['root'], 'etc/xapi.conf.d/ssh-auto-mode.conf.extra')
```
rosslagerwall, line 1292: Does this file also need to be saved/restored on upgrade?

backend.py
```diff
@@ -1280,6 +1281,36 @@ def writeFstab(mounts, primary_disk, logs_partnum, swap_partnum, disk_label_suff
...
+            f.write('ssh-auto-mode=true\n')
+        logger.log("SSH auto-mode enabled")
+    else:
+        with open(ssh_auto_mode_conf_path, 'w') as f:
+            f.write('ssh-auto-mode=false\n')
+
+        # Enable or disable SSH service based on mode
+        if ssh_mode == 'on':
+            util.runCmd2(['chroot', mounts['root'], 'systemctl', 'enable', 'sshd.service'])
+            util.runCmd2(['chroot', mounts['root'], 'systemctl', 'start', 'sshd.service'])
```
rosslagerwall, line 1307: I don't think starting or stopping the service makes sense in the install environment.

backend.py
LunfanZhang: It is only server for fresh install, as for upgrade scenario, it depends on the configuration in XAPI db, generally, XS9 will inherit the ssh-auto-mode in XS8.

LunfanZhang: We place it under `/etc/xapi.conf.d/` because XAPI automatically initializes all configuration files and extracts all key-value pairs from this dir.

backend.py
freddy77: Yes, this could be a problem, beside the mentioned conflict between installer ssh and system ssh launching a daemon could cause issue when unmounting file systems. Probably as systemd is running outside the chroot this is probably going to fail anyway.

backend.py
rosslagerwall: Is there any reason it has a `.extra` extension rather than a `.conf` extension like the other files in `xapi.conf.d`?

rosslagerwall: IIUC after a pool eject, the pool will use the value configured here. If you do an upgrade followed by a pool eject, because the this config is not kept it will revert to the default value rather than the value originally configured at install time. Does that matter?

LunfanZhang: Adding a `extra` extension is due to the XAPI will generate the `ssh-auto-mode.conf` in XS8 as well, although we only would like to merge this to XS9, assume someday some request from user would like this functionality when we backport the it to XS8, we can ensure the configuration generate here will not override by XAPI, and it is some hack here as I using the XAPI initializes order, xapi will initializes the k-v pair base on the file name, which means the configure here take effect but not the one geneated by XAPI

LunfanZhang: And If user do an upgrade followed by a pool eject, the host will apply the default value on XS9 level which will then enable auto-mode, I tend to think it also make sense, it will not bother the user. if user do not want auto-mode, they could turn it off manually as well.
If we prefer to keep the original setting, I could raise another PR/Commits for do that as well.

PR #289, opened 2025-08-01
CP-308408: force use of OVS in XS 9

MarkSymsCtx:


answerfile.py
```diff
@@ -193,13 +193,8 @@ def parseCommon(self):
...
         results.update(self.parseSource())
 
-        nodes = getElementsByTagName(self.top_node, ['network-backend'])
-        if len(nodes) > 0:
-            network_backend = getText(nodes[0])
-            if network_backend == NETWORK_BACKEND_BRIDGE:
-                results['network-backend'] = NETWORK_BACKEND_BRIDGE
-            elif network_backend in [NETWORK_BACKEND_VSWITCH, NETWORK_BACKEND_VSWITCH_ALT]:
-                results['network-backend'] = NETWORK_BACKEND_VSWITCH
+        # OVS is the only supported network backend regardless of the answerfile contents
```
rosslagerwall, line 196: The network backend is already set to `NETWORK_BACKEND_DEFAULT` aka `NETWORK_BACKEND_VSWITCH` in line 98 of install.py so I don't think you need to explicitly set it here or in product.py.

PR #290, opened 2025-08-13
CA-413099: Preserve /var/lib/xcp/verify-certificates

liulinC:
This commit has two fixes:
- Preserve /var/lib/xcp instead of /var/xapi /var/xapi is a symbol link to /var/lib/xcp, however, the symbol link is broken during RPU, due to temp mount. Thus, the real folder should be used instead of link
- rename verify_certificates to verify-certificates, as it is the real file in use


upgrade.py
```diff
@@ -465,7 +464,8 @@ def buildRestoreList(self, src_base):
         # CA-195388: Preserve /etc/mdadm.conf across upgrades
         restore_list += ['etc/mdadm.conf']
 
-        restore_list += ['var/lib/xcp/verify_certificates']
+        # CA-413099: Preserve /var/xapi/verify-certificates for certificates checking enable/disable status
```
rosslagerwall, line 467: `/var/lib/xcp/verify-certificates` ?

rosslagerwall:
> Preserve /var/lib/xcp instead of /var/xapi /var/xapi is a symbol link to /var/lib/xcp, however, the symbol link is broken during RPU, due to temp mount. Thus, the real folder should be used instead of link

This is why relative symlinks should normally be preferred over absolute symlinks...

upgrade.py
liulinC: they actually refer to the same file, as symbol link as my commit message.
but good to update to avoid confusing.

rosslagerwall (approved):

PR #292, opened 2025-08-29
CA-416373: virtual interface name exceed the limitation

liulinC:
Linux kernel limit the interface name length to 16(exclude), refer to dev_valid_name for the details.

The vlan interface name was defined as `host_interface.vlanID`, this was fine when the interface was defined as ethX, but is a problem when systemd provide a long interface name for the host interface.

To resolve the issue, the vlan interface name is normalized to ensure it does not exceed the limitation.

The interface bring up before `networkctl reload` is removed, as `networkctl reload` will bring it up anyway

### DIFF
--- a/netinterface.py
+++ b/netinterface.py
@@ -5,7 +5,10 @@ import configparser
 
 import util
 import netutil
+import random
+import string
 from xcp import logger
+import constants
 
 def getText(nodelist):
     rc = ""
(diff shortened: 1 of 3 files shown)


netinterface.py
```diff
@@ -107,7 +114,18 @@ def get(self, name, default=None):
...
+
+        # linux kernel has limitation for the interface name length, 1 for the "." in the middle
+        max_host_interface_length = constants.INTERFACE_NAME_LENGTH_MAX - constants.VLAN_LENGTH_MAX - 1;
+        if len(iface) <= max_host_interface_length:
+            return f"{iface}.{self.vlan}"
+        else:
+            name = f"{random_string(max_host_interface_length)}.{self.vlan}"
+            logger.log(f"Generated {name} for {iface}.{self.vlan}")
+            return name
+
```
changlei-li, line 128: getInterfaceName is called in three functions:
- writeSystemdNetworkdConfig
- waitUntilUp
- requireNetworking

I think there are two usages of this function
- If with vlan, create a new interface name (writeSystemdNetworkdConfig)
- If with vlan, get the created name (waitUntilUp)

For introducing random string, the second usage is broken

liulinC: it is NOT a problem here, as `waitUntilUp` never called.  (it was used in intsysv), 
systemd would bring up the interface automatically.
However, we do need to check the tui/network.py file. 
I will re-consider this.

PR #298, opened 2025-09-15
CA-416618: Support degraded SW RAID installations

GeraldEV:
During recovery a user may choose to re-install on a single physical disk while the other part of the SW RAID device is in-operable. To achieve this the answerfile must accept a "missing" second physical disk.

Performing upgrade & restore operations over a degraded device is not supported - this use case was requested by the internal customer as part of their reset process; in which a fresh installation will be performed on a single target device

### DIFF
--- a/answerfile.py
+++ b/answerfile.py
@@ -317,11 +317,14 @@ class Answerfile:
         results['swraid'] = getBoolAttribute(node, ['swraid'], default=False)
 
         if results['swraid']:
-            results['primary-disk'] = ""  # Populated with disk-label-suffix during installer prep
-            results['physical-disks'] = [normalize_disk(disk) for disk in getText(node).split(",")]
-
-            if len(results['physical-disks']) != 2:
+            disks = getText(node).split(",")
+            if len(disks) != 2:
                 raise AnswerfileException("swraid primary-disk requires exactly two physical disks")
+            if disks.count("missing") > 1:
+                raise AnswerfileException("swraid primary-disk supports only one missing disk")
+
+            results['primary-disk'] = ""  # Populated with disk-label-suffix during installer prep
+            results['physical-disks'] = [normalize_disk(disk) for disk in disks if disk != "missing"]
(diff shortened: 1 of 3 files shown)


rosslagerwall (dismissed):

stormi:
Note: in our old PR regarding the handling of existing soft raids, still open, we would delay assembling the raid until the user decides whether they want to install on the existing raid or on a single disk. I think this would have covered this use case.

doc/answerfile.txt
```diff
@@ -198,6 +198,8 @@ Format of 'source' and 'driver-source'
         of the primary-disk element. Devices should be comma separated.
         The first device will be treated as the original, the second device will
         mirror the first.
+        The second device may be "missing" to proceed with a degraded RAID
```
freddy77, line 201: This is not exactly what the code is doing, the code is ignoring one `missing`, first or second

answerfile.py
```diff
@@ -317,11 +317,14 @@ def parseDisks(self):
...
 
         if results['swraid']:
-            results['primary-disk'] = ""  # Populated with disk-label-suffix during installer prep
-            results['physical-disks'] = [normalize_disk(disk) for disk in getText(node).split(",")]
-
-            if len(results['physical-disks']) != 2:
+            disks = getText(node).split(",")
+            if len(disks) != 2:
                 raise AnswerfileException("swraid primary-disk requires exactly two physical disks")
+            if disks.count("missing") > 1:
```
freddy77, line 323: From the documentation below maybe it should be
```python
if disks[1] == "missing":
    disks = disks[:1]
```

GeraldEV: I think it is better to show an error that the user has specified too many "missing" disks rather than a cryptic error that it couldn't "normalise 'missing'"

freddy77: With the updated documentation I think it's fine

freddy77 (approved):

PR #300, opened 2025-09-22
CA-416806: Turn off srso mitigation for performance

liulinC:


snwoods (approved):
I can't comment on the security aspect, but approving for the performance impact. The largest absolute regression drops from 171% to 32% with this patch.

PR #302, opened 2025-09-25
CA-417641 Update chrony-dhcp server file path

changlei-li:
On XS8, the chrony add dhcp_ntp_servers to
`/var/lib/dhclient/chrony.servers.$interface` then call `/usr/libexec/chrony-helper update-daemon`
On Xs8, the chrony add dhcp_ntp_servers to
`/run/chrony-dhcp/$interface.sources` and reload chrony.

In this PR, host-installer check the chrony-dhcp server file to determine the ntp mode. We can check chrony-helper to see which path is used.


product.py
```diff
@@ -426,7 +426,12 @@ def _check_dhcp_ntp_status(self):
             # chrony.sh not executable indicates not using DHCP for NTP
             return False
 
-        for f in glob.glob(self.join_state_path('var/lib/dhclient/chrony.servers.*')):
+        if os.path.exists(self.join_state_path('usr/libexec/chrony-helper')):
+            chrony_dhcp_server_path = 'var/lib/dhclient/chrony.servers.*'
```
rosslagerwall, line 430: The host-installer `master` branch is only for XS9 so this part isn't necessary.

changlei-li: I think it is upgrade case, maybe upgrade from XS8

rosslagerwall: Oh, right.

Rather than using the presence of `chrony-helper` to guess where the state is stored, I think it could be simplified a bit by just globbing both paths. E.g.

```
for f in glob.glob(self.join_state_path('var/lib/dhclient/chrony.servers.*')) + glob.glob(self.join_state_path('run/chrony-dhcp/*.sources')):
```

changlei-li:
Ah, I am doubting that if the run dir can be reserved on the mount point now. Generally, `/run` dir will be refreshed after reboot. Let me confirm.

product.py
changlei-li: I tested that the `run/chrony-dhcp` dir didn't exist on mount point. I remove the check for the files, just use execute permission of chrony.sh to determine if dhcp mode is used for ntp. See detail in the comment below.

stephenchengCloud:
It makes sense to me to check the execute permission of chrony.sh according to https://github.com/xenserver/host-installer/blob/v11.0.34/backend.py#L520

stephenchengCloud (approved):

liulinC (dismissed):

stephenchengCloud (approved):

DeliZhangX (approved):

PR #304, opened 2025-09-28
CA-415170: Set find_multipaths to greedy

DeliZhangX:
In XS8 host installer sets "find_multipaths" to "no", but the XS_MPATH_CONF file (/etc/multipath.conf) of sm is already deprecated in XS9.
So here just directly generate its specific config file /etc/multipath/conf.d/installer.conf:
```
defaults {
    find_multipaths greedy
}
```
where "greedy" is equivalent to "no" in newer version DM-Multipath.


DeliZhangX:
Will be replaced by other solution.

PR #332, opened 2026-06-15
CP-312957: add a new parameter for LLDP blocking list

xueqingz:


MarkSymsCtx (approved):

backend.py
```diff
@@ -1444,6 +1445,18 @@ def writeResolvConf(mounts, hn_conf, ns_conf):
                 resolvconf.write("nameserver %s\n" % ns)
         resolvconf.close()
 
+def writeLLDPBlocklist(mounts, lldp_block_drivers):
+    if not lldp_block_drivers:
+        return
+    seen = set()
+    drivers = [d for d in lldp_block_drivers if not (d in seen or seen.add(d))]
```
liulinC: This looks complicated, does the order matter, 
- if not, can use `list(set(lldp_block_drivers))` ?
- if matters, can use `list(dict.fromkeys(lldp_block_drivers))`

backend.py
```diff
@@ -188,6 +188,7 @@ def getFinalisationSequence(ans):
         Task(writeFstab, A(ans, 'mounts', 'primary-disk', 'logs-partnum', 'swap-partnum', 'disk-label-suffix', 'fs-type'), []),
         Task(enableAgent, A(ans, 'mounts', 'network-backend', 'services'), []),
         Task(configureSSHMode, A(ans, 'mounts', 'ssh-mode'), []),
+        Task(writeLLDPBlocklist, A(ans, 'mounts', 'lldp-block-drivers'), []),
```
liulinC: The upgrade would override the cmd options, it this by design? or do we need log for this?

install.py
```diff
@@ -131,6 +132,10 @@ def go(ui, args, answerfile_address, answerfile_script):
             constants.CC_PREPARATIONS = True
         elif opt == "--mount":
             disktools.DeviceMounter.addMountPoints(val)
+        elif opt == "--lldp_block_driver":
```
liulinC, line 135: This is only available from cmdline, no UI change, 
Presume this is the only way supported.

xueqingz: yes

backend.py
liulinC: I mean the restored file takes priority than cmd line.  Is this by design?

xueqingz: Ok, I understand what you mean. I think it's by design, The design said “These drivers persist in the installed XenServer host and remain after RPU.” Passing `lldp_block_driver=` during an upgrade seems not a expected flow, @minglumlu Could you confirm here ?

minglumlu: It would be a rare case that a user wants to add block drivers during upgrade. But it's harmless to get the new drivers merged with the existing one.

liulinC (approved):

PR #336, opened 2026-06-23
CP-311882: Remove safe mode boot option

GeraldEV:
Safe mode is generally untested and not used, if a customer experiences issues they should contact support rather that attempting to use safe mode.

### DIFF
--- a/backend.py
+++ b/backend.py
@@ -1053,8 +1053,7 @@ def prepFallback(mounts, primary_disk, primary_partnum):
 
 def buildBootLoaderMenu(mounts, xen_version, xen_kernel_version, boot_config, serial, boot_serial, host_config, primary_disk, disk_label_suffix):
     short_version = kernelShortVersion(xen_kernel_version)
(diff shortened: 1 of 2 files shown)


backend.py
```diff
@@ -1055,9 +1055,6 @@ def buildBootLoaderMenu(mounts, xen_version, xen_kernel_version, boot_config, se
     short_version = kernelShortVersion(xen_kernel_version)
     common_xen_params = "dom0_mem=%dM,max:%dM" % ((host_config['dom0-mem'],) * 2)
     common_xen_unsafe_params = "watchdog dom0_max_vcpus=1-%d" % host_config['dom0-vcpus']
```
rosslagerwall: It doesn't make sense to separate out unsafe params now if there is no safe mode.

rosslagerwall (approved):

KevinLCtx (approved):

freddy77 (approved):

PR #337, opened 2026-08-11
CP-313559 Support grub-mkconfig action in installer (test only)

TimSmithCtx:
Provide for a transition to using grub-mkconfig for generating the bootloader configuration by writing information required for an update-grub action into a configuration file and invoking it, but only if the install actually has the update-grub action.

The update-grub action is expected to honor XS__GRUB_TEST_ONLY and write only a test configuration.

A later change will detect that the install does not want test mode, and both leave out the XS__GRUB_TEST_ONLY setting, and avoid writing the legacy bootloader configuration, expecting the update-grub action to have done it for real.


backend.py
```diff
@@ -1087,25 +1109,55 @@ def buildBootLoaderMenu(mounts, xen_version, xen_kernel_version, boot_config, se
...
 
+    # Write an xs_settings file to the root partition, but only if
+    # the update-grub action is available in the install
+    if os.path.exists(os.path.join(mounts['root'], '/usr/sbin/update-grub')):
+        xs_settings = os.path.join(mounts['root'], 'etc/grub.d/xs_settings')
+        with open(xs_settings, "w") as f:
+            boot_default = 1 if boot_serial else 0
+            serial_id = serial.id if serial else 0
+            serial_baud = serial.baud if serial else 115200
+            root_label=constants.rootfs_label%disk_label_suffix
```
GeraldEV: Inconsistent spaces with the rest of the file/changes

GeraldEV (approved):
Looks good, I had one or two minor comments but I think they're all just artifacts of the old implementation and mostly non-functional so I'll do them myself in another PR at some point

Only question is; is there a better phrasing of "safe mode"? I know your comment explains the difference but it won't be explained in the xs_settings file

TimSmithCtx:
> Only question is; is there a better phrasing of "safe mode"? I know your comment explains the difference but it won't be explained in the xs_settings file

I'm open to suggestions. I inherited the terminology from a starting point which pre-dated the removal of safe mode and just kept it because I couldn't think of a replacement which worked.

Possibly the correct place to expand on this would be the grub-mkconfig fragment which handles the contents of the xs_settings file?

rosslagerwall (approved):

